PR #4, opened 2022-06-01
chore: tell gorilla what the `resource.{requests,limits}.memory` is

minyoung:
Provide gorilla its memory request and limit. This way, gorilla can use this information to mark itself unhealthy for k8s to stop sending it requests: https://github.com/wandb/core/pull/9695

Note: this functionality of marking itself unhealthy based on memory usage is turned **off** by default.


dacbd:
seems implemented already

PR #6, opened 2023-04-07
Agent v0

bcsherma:
This PR contains a v0 of a helm chart for deploying launch agents in k8s. More details in the README.md

One note: there is a link in the README.md to docs.wandb.ai/guides/launch/agent which currently goes nowhere but will point to our agent docs once this docs pr gets merged https://github.com/wandb/docodile/pull/151

The chart creates a wandb namespace, a new service account in that namespace, and some roles that are bound to that service account so the agent can create and monitor builds and runs. It also creates a deployment for the agent and a secret that contains the agent's wandb api key that we mount as an environment variable in the agent container.


charts/launch-agent/README.md
```diff
@@ -0,0 +1,15 @@
...
+The launch agent is a Kubernetes Deployment that runs a container that connects to the W&B API and watches for new runs in one or more launch queues. When the agent pops a run off the queue(s), it will launch a Kubernetes Job to execute the run on the W&B user's behalf.
+
+To deploy an agent, you will need to specify the following values:
+* `agent.apiKey`: Your W&B API key
+* `launchConfig`: The literal contents of a launch agent config file that will be used to configure the agent. See the [launch agent docs](https://docs.wandb.ai/guides/launch/agent) for more information.
+
+You can provide these values by modifying the contents of `values.yaml` or by passing them in as command line arguments to `helm install`, e.g.
+
+```bash
+helm install <package-name> <launch-agent-chart-path> --set agent.apiKey=<your-api-key> --set launchConfig="$(cat <path-to-launch-config.yaml>)"
```
vanpelt, line 14: I think you can use `--set-file={path}` instead of cat'ing the file

charts/launch-agent/templates/rbac.yaml
```diff
@@ -0,0 +1,62 @@
...
+  kind: Role
+  name: wandb-launch-agent
+  apiGroup: rbac.authorization.k8s.io
+---
+# role binding to create ML jobs in another namespace (could use cluster role binding if we want to launch cluster wide)
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: wandb-launch-cluster-role-binding
+  namespace: default #TODO: SET YOUR TRAINING NAMESPACE
```
vanpelt, line 54: You should probably wire this up to variables.

charts/launch-agent/templates/volcano.yaml
```diff
@@ -0,0 +1,27 @@
...
+  template:
+    spec:
+      containers:
+        - name: volcano-install-job
+          image: bitnami/kubectl:latest
+          command:
+            - "/bin/sh"
+            - "-c"
+            - |
+              kubectl apply -f https://raw.githubusercontent.com/volcano-sh/volcano/master/installer/volcano-development.yaml
```
vanpelt, line 19: Curious what volcano is?

charts/launch-agent/values.yaml
```diff
@@ -0,0 +1,15 @@
...
+  apiKey: ""
+  # Container image to use for the agent.
+  image: wandb/launch-agent-dev:latest
+  # Image pull policy for agent image.
+  imagePullPolicy: Always
+  # Resources block for the agent spec.
+  resources:
+    limits:
+      cpu: 1000m
+      memory: 1Gi
```
vanpelt, line 12: We should probably tell users to override this in our README as it's won't make sense for the majority of workloads.

vanpelt (approved):

charts/launch-agent/templates/volcano.yaml
bcsherma: It's a batch scheduler that is becoming widely used for distributed trainings on kubernetes https://volcano.sh/en/

bcsherma: We are relying on this to support launching multi node distributed training in k8s

bcsherma: Unfortunately they do not have a living helm repo.

PR #9, opened 2023-05-26
add variable for mounting git credentials

bcsherma:
This PR introduces a `gitCreds` variable which can be set to the literal contents of a `.git-credentials` file. The chart puts this into a secret and mounts it in the home directory of the agent container, if `gitCreds` is specified by the user. It will also mount
```
[credentials]
helper = store
```
in `~/.gitconfig` to prevent git from prompting for a username and password and instead going straight to the cred store we have mounted.

I also added a table that documents all the variables to the readme.


KyleGoyette (approved):
GIT CREDS 🔥

PR #25, opened 2023-08-15
Operator W&B Chart

jsbroks:


charts/operator-wandb/charts/app/Chart.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: v2
+name: app
+type: application
+description: A Helm chart for Kubernetes
+
+version: 0.1.0
+appVersion: "0.33.0"
```
gls4, line 7: Why are we using this version of the app here?

jsbroks: app version doesn't really mean anything here, we can set it as 1.0.0?

gls4 (approved):

PR #27, opened 2023-08-24
Introduce  property to avoid secret generation

rpemsel:
Introduces a property that prohibits secret generation in case an externally created secret is meant to be used.


rpemsel:
Lint complains that the App Version is top old :(. So I updated IT to 0.40.0

flamarion:
I sent a commit to your PR @rpemsel that might fix the lint.

PR #28, opened 2023-08-26
Avoid secrets generation

flamarion:


flamarion:
This change may force people to provide a persistent volume claim or external object storage and include PR #27 from Mercedes.
If they don't provide, they will lose data, but the deployment will not fail waiting for a Volume Claim that will never happen without a previous PV.

gls4 (approved):
I think this is fine as long as we make sure to note that customers need to make a PV or run the risk of losing data.

jsbroks (approved):

PR #29, opened 2023-09-05
feat: Update README, add CODEOWNERS

gls4:
Amendments made to the wandb chart README, and added CODEOWNERS file.


flamarion (approved):

flamarion:
@gls4, you need to bump the chart version [here](https://github.com/wandb/helm-charts/blob/main/charts/wandb/Chart.yaml#L5)

PR #33, opened 2023-10-03
feat: Adding support for k8s webhooks secret store

wandb-zacharyblasczyk:
This PR adds a role and role binding to the existing wandb Service Account. 

One thing that needs to be determined is if we want these to be optional as the existing Service Account Implementation is optional. 

* Bump to the W&B server version.


charts/wandb/templates/role.yaml
```diff
@@ -0,0 +1,15 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: secret-manager-role
```
jsbroks, line 5: why is this static

jsbroks (changes requested):

charts/wandb/templates/role.yaml
wandb-zacharyblasczyk: This can be whatever pattern. Whatever name you want.

charts/wandb/templates/role.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+  {{- include "wandb.labels" . | nindent 4 }}
+rules:
+  - apiGroups: [""]
+    resources: ["secrets"]
+    verbs: ["get", "create", "update", "delete"]
+  - apiGroups: [""]
+    resources: ["namespaces"]
+    verbs: ["get"]
```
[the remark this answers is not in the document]
wandb-zacharyblasczyk: It does need the namespace right now. 

We can remove it but, there is a namespace check that happens on the server side as of 0.44.0+. 
There would need to be a server side change to remove this.  
This is for backwards compatibility.

charts/wandb/templates/role.yaml
jsbroks: Everything should be based off the chart name

wandb-zacharyblasczyk: Done.

charts/wandb/templates/rolebinding.yaml
```diff
@@ -0,0 +1,15 @@
...
+  labels:
+  {{- include "wandb.labels" . | nindent 4 }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "wandb.serviceAccountName" . }}
+roleRef:
+  kind: Role
+  name: {{ include "wandb.secretManagerRoleName" . }}
+  apiGroup: rbac.authorization.k8s.io
+...
```
jsbroks, line 15: whats this?

jsbroks (approved):

wandb-zacharyblasczyk:
feat: Update wandb Helm Chart to support kubernetes secret store

PR #35, opened 2023-10-06
fix: Add app permissions to create secrets

jsbroks:


charts/operator-wandb/charts/app/templates/role.yaml
```diff
@@ -0,0 +1,20 @@
...
+    {{- include "app.commonLabels" . | nindent 4 }}
+    {{- include "app.labels" . | nindent 4 }}
+    {{- if .Values.clusterRole.labels -}}
+    {{-   toYaml .Values.clusterRole.labels | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- if .Values.role.annotations -}}
+    {{-   toYaml .Values.role.annotations | nindent 4 }}
+    {{- end }}
+rules:
```
wandb-zacharyblasczyk, line 17: We should modify role to include this rule:
  - apiGroups: [""]
    resources: ["namespaces"]
    verbs: ["get"]
We add this to be able to check for existence of (but not create) the namespace

wandb-zacharyblasczyk (approved):

charts/operator-wandb/charts/app/templates/role.yaml
jsbroks: why are we not deploying into the namespace the pod is deploy into?

wandb-zacharyblasczyk: We are that role is scoped to the namespace. @andrewtruong can clarify why this is needed.

PR #39, opened 2023-10-31
docs(launch): add example agent config to values.yaml

bcsherma:


charts/launch-agent/values.yaml
```diff
@@ -27,6 +27,12 @@ additionalTargetNamespaces:
 
 # This should be set to the literal contents of your launch agent config.
 launchConfig: |
+  queues:  # List of queues to poll for jobs.
+  - <YOUR QUEUE NAME>
```
KyleGoyette, line 31: let's make this lower case

KyleGoyette (approved):
One nit

PR #42, opened 2023-11-01
fix: Fix bucket queue env var

flamarion:
fix: Fix bucket queue env var

### DIFF
--- a/charts/wandb/templates/deployment.yaml
+++ b/charts/wandb/templates/deployment.yaml
@@ -115,7 +115,7 @@ spec:
             - name: AWS_S3_KMS_ID
               value: {{ .Values.bucketKmsKeyArn }}
             - name: BUCKET_QUEUE
-              value: {{ default "internal://" .Values.bucketQueue }}
+              value: {{ .Values.bucketQueue | default "internal://" }}
             {{- end }}
             - name: MYSQL
               value: {{ "mysql://$(DB_USER):$(DB_PASSWORD)@$(DB_HOST)/$(DB)" }}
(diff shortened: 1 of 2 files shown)


charts/wandb/templates/deployment.yaml
```diff
@@ -115,7 +115,7 @@ spec:
             - name: AWS_S3_KMS_ID
               value: {{ .Values.bucketKmsKeyArn }}
             - name: BUCKET_QUEUE
-              value: {{ default "internal://" .Values.bucketQueue }}
+              value: {{ .Values.bucketQueue | default "internal://" }}
```
vanpelt, line 118: @flamarion you sure this is needed?  According [to the docs](https://helm.sh/docs/chart_template_guide/function_list/#default) the other syntax is also fine.

vanpelt (approved):

charts/wandb/templates/deployment.yaml
flamarion: [Both](https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function) are valid @vanpelt, but the initial approach didn't work for one of our customers.

Maybe the reason is that we need to set it `bucketQueue` explicitly  with `""` in the template based on this quote from the description: `The definition of "empty" depends on type:`

This is the full context.
```
In the above, if .Bar evaluates to a non-empty value, it will be used. But if it is empty, foo will be returned instead.

The definition of "empty" depends on type:

Numeric: 0
String: ""
Lists: []
Dicts: {}
Boolean: false
And always nil (aka null)
```

I will attempt it before merging.

PR #43, opened 2023-11-10
feat: Add support onprem storage provider

flamarion:


flamarion:
I tested locally using the helm chart directly instead of the Operator because I didn't know how to make the operator use my branch to pull the modifications. 
So basically, I used the same file that I added to the example to override the values and the local helm chart for `operator-wandb` to deploy the W&B from my branch.

```
 helm upgrade --install wandb ../../src/helm-charts/charts/operator-wandb -f wandb-values.yaml
 ```

PR #47, opened 2023-11-22
Launch: Support multiple instances in the same namespace

KyleGoyette:
Also fixes: https://wandb.atlassian.net/browse/WB-16022


bcsherma (approved):
Lgtm!

PR #48, opened 2023-12-01
disable otel install

gls4:
Disabling install of the bundled OTEL collector for now.


nfoucha (approved):

wandb-zacharyblasczyk (commented):
I don't think we want to make this the default. (maybe we do)

I have set it here for now:
https://github.com/wandb/terraform-aws-wandb/pull/154/files#diff-dc46acf24afd63ef8c556b77c126ccc6e578bc87e3aa09a931f33d9bf2532fbbR226

PR #53, opened 2023-12-12
chore: Remove statd host port if otel is not enabled

yogeshg:
- chore: Remove statd host port if otel is not enabled
- empty

### DIFF
--- a/charts/operator-wandb/charts/app/templates/deployment.yaml
+++ b/charts/operator-wandb/charts/app/templates/deployment.yaml
@@ -156,15 +156,15 @@ spec:
 
             - name: GORILLA_SESSION_LENGTH
               value: "{{ .Values.global.auth.sessionLengthHours }}h"
-            
+
+            {{- if and .Values.global .Values.global.observability }}
             {{- if eq (default "custom" .Values.global.observability.mode) "otel" }}
             - name: GORILLA_STATSD_PORT
               value: "8125"
             - name: GORILLA_STATSD_HOST
               value: "0.0.0.0"
             {{- end }}
-            - name: FOO_DEC_13
-              value: "BAR_DEC_13"
+            {{- end }}
 
             - name: BUCKET
               value: "{{ include "app.bucket" . }}"
--- a/charts/operator-wandb/charts/app/templates/deployment.yaml
+++ b/charts/operator-wandb/charts/app/templates/deployment.yaml
@@ -157,12 +157,12 @@ spec:
             - name: GORILLA_SESSION_LENGTH
               value: "{{ .Values.global.auth.sessionLengthHours }}h"
             
-            {{ if .Values.otel.install }}
+            {{- if eq .Values.global.observability.mode "otel" }}
             - name: GORILLA_STATSD_PORT
-              value: "8125"
-            - name: GORILLA_STATSD_HOST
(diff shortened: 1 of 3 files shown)


jsbroks (approved):

charts/operator-wandb/charts/app/templates/deployment.yaml
```diff
@@ -157,10 +157,12 @@ spec:
             - name: GORILLA_SESSION_LENGTH
               value: "{{ .Values.global.auth.sessionLengthHours }}h"
             
+            {{ if .Values.otel.install }}
```
jsbroks, line 160: you cannot access root values in sub charts. You would need to use `globals`, helm automically scopes the values based on sub paths

jsbroks: **gpt to explain what I mean:**


In Helm, a chart can include other charts, which are referred to as subcharts. Each chart has its own `values.yaml` file which defines the configurable parameters and their default values. When you install or upgrade a chart, you can provide custom values to override these defaults. These values form a hierarchy that reflects the structure of the chart and its subcharts.

Here's a breakdown of each part of your statement:

1. "you cannot access root values in sub charts":
   
   This means that values defined in the `values.yaml` file at the root of the parent chart are not directly accessible inside the templates of a subchart. You usually define values specific to a subchart within its own `values.yaml` file.

2. "You would need to use globals":

   Helm has a concept of global values, which are specified under the `global` key in the `values.yaml` file. These global values are accessible to all charts and subcharts. If you need to pass a value from the parent chart to a subchart and make it accessible to the subchart's templates, you can define this value under the `global` key in the parent's `values.yaml` file.

3. "helm automatically scopes the values based on sub paths":

   Helm automatically understands the structure of your chart and its subcharts and scopes the values accordingly. This scoping is based on the directory paths within the chart. Each subchart gets its values from its `values.yaml` file and any values passed to the parent chart that are scoped for that subchart. This helps to avoid naming conflicts and keeps the values organized according to the chart hierarchy.

In practice, when you want a subchart to use a value from the parent chart, you essentially have two choices:

- Define a global value that can be accessed by both the parent chart and all of its subcharts.
- Override subchart's values from the parent chart by specifying the subchart's name as a prefix to the keys when you pass values that should apply specifically to the subchart.

Here's a simple example of a global value:

```yaml
# Inside the parent chart's values.yaml
global:
  appVersion: "1.0.0"
```

This `appVersion` value would be accessible to the parent chart and all its subcharts.

And here's how you might override a value for a subchart from the parent chart's custom values (usually passed via the `-f` flag or `--set` flag during `helm install` or `helm upgrade`):

```yaml
# Custom values passed to override subchart values
subchartName:
  someParameter: "Overridden Value"
```

In this case, `someParameter` is a value defined in the `subchartName` subchart, and you're specifying a new value for it from the context of the parent chart.

jsbroks:
Im not sure if there is an easy way to do this without making the logic more complciated. it might be better to leave the port open as it doesn't do much

charts/operator-wandb/charts/app/templates/deployment.yaml
yogeshg: used global. tested on my dev env.

PR #58, opened 2024-01-10
fix: Auth conflict

wandb-zacharyblasczyk:


jsbroks (approved):

wandb-zacharyblasczyk:
Only for backwards compatibility.

PR #59, opened 2024-01-10
feat: Create api key secret in all namespaces

TimH98:
This is a companion PR to [this SDK PR](https://github.com/wandb/wandb/pull/6722). Instead of only creating the `wandb-api-key` secret in the agent's namespace, create one in every namespace used by the agent. This way, secrets are created when the agent is spun up instead of at job runtime. This also allows the helm deployment to manage cleanup of the secrets when the agent is spun down.

Since we allow multiple agents on the same namespace, and different agents may want to use different secrets, these separate secrets are denoted by the agent's `.Release.Name`, which is passed to the agent through the `WANDB_RELEASE_NAME` env var.

### DIFF
--- a/charts/launch-agent/templates/secret.yaml
+++ b/charts/launch-agent/templates/secret.yaml
@@ -8,6 +8,8 @@ kind: Secret
 metadata:
   name: wandb-api-key-{{ $root.Release.Name }}
   namespace: {{ $ns }}
+  labels:
+    wandb.ai/created-by: helm-chart
 type: kubernetes.io/basic-auth
 stringData:
   password: {{ required "Please set agent.apiKey to a W&B API key" $root.Values.agent.apiKey }}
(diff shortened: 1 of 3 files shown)


charts/launch-agent/templates/secret.yaml
```diff
@@ -1,16 +1,20 @@
...
+{{- $unique := uniq (append .Values.additionalTargetNamespaces .Values.namespace) }}
+{{- range $ns := $unique }}
 ---
 apiVersion: v1
 kind: Secret
 metadata:
-  name: wandb-api-key-{{ .Release.Name }}
-  namespace: {{ .Values.namespace }}
+  name: wandb-api-key-{{ $root.Release.Name }}
+  namespace: {{ $ns }}
```
nickpenaranda, line 10: let's add a label here: `wandb.ai/created-by: helm-chart` so we can distinguish it from an agent-installed secret

similarly, when the agent creates the secret let's set a different label: `wandb.ai/created-by: launch-agent`

this gives us the ability to positively identify these resources were created by us

bcsherma: +1, great idea

nickpenaranda (approved):

PR #61, opened 2024-01-16
fix: Enable weave by default

zacharyblasczyk:


jsbroks (approved):

yogeshg (changes requested):
We don't need this PR until we're ready to rollout weave by default.

dacbd:
stale

PR #76, opened 2024-02-07
fix: Allow ALB time to register new pod ip for the wandb app and console pod

wandb-zacharyblasczyk:
Here is a description of the issue and workaround. 
https://github.com/kubernetes-sigs/aws-load-balancer-controller/issues/1064


wandb-zacharyblasczyk:
This fixes the issue in AWS with the AWS Ingress Controller, I am not sure if there is a similar issue in AKS or GKE, but if there is this should help and is a good starting place to make tweaks.

nfoucha (approved):

PR #78, opened 2024-02-12
chore(launch): bump agent version to 0.16.3

bcsherma:
bumping launch agent version to 0.16.3 post release


TimH98 (approved):
Stampy stamp

PR #82, opened 2024-02-22
kafka

jsbroks:


charts/operator-wandb/charts/parquet/templates/deployment.yaml
```diff
@@ -57,10 +57,19 @@ spec:
           env:
             - name: ONLY_SERVICE
               value: gorilla-parquet
-
             - name: LOGGING_ENABLED
               value: 'true'
 
+            - name: KAFKA_BROKERS
```
yogeshg, line 63: prolly gonna go in frfu sts.

charts/operator-wandb/values.yaml
```diff
@@ -173,3 +184,20 @@ otel:
...
+      password: "inter-broker"
+    controller:
+      password: "controller-pw"
+    # The client usernames and passwords are alias references from the global values section at the beginning of this file.
+    # Please update the values there to ensure proper propogation to the application
+    client:
+      users: *kafkaUsers
+      passwords: *kafkaPasswords
+  kraft:
+    # This field is a UUID. It is *strongly* recommended to supply passwords yourself for production installs.
```
yogeshg, line 202: **nitpick (non-blocking):** ```suggestion
    # This field is a UUID. It is *strongly* recommended to supply a new UUID yourself for production installs.
```

**Review by yogeshg — APPROVED** [review]

File: charts/operator-wandb/values.yaml [hunk 3]
```diff
@@ -74,6 +74,17 @@ global:
     parameters: {}
     caCert: ""
 
+  kafka:
+    # The following values are anchored here, and referenced by alias later for the kafka dependency chart.
+    # List of username for kafka, first being the admin user.  Must have the same number of elements as the passwords value
+    users: &kafkaUsers
```
yogeshg, line 80: Can we make this kafkaUser instead of kafkaUsers?
Reason being that this is going to be used by frfu, and there we'll need a variable with a single string that contains the user name.
Here we'll need:
```
users: &kafkaUser
passwords: &kafkaPassword "wandb"
```
Below we'll need:
```
      users:
            - *kafkaUser
      passwords: *kafkaPassword
```

nfoucha: Is this necessary or can frfu be aware that it's getting a list and a comma delimited string?  This is also at the global layer so we can always add a pared down "individual" user/password value to the app specific blocks.

PR #84, opened 2024-02-26
feat(launch): support kaniko builds with user provided pvc and dockerconfig

bcsherma:
This PR enables the deployment of an agent capable of building images with kaniko on coreweave kubernetes.
 
- Adds `kanikoPvcName` and `kanikoDockerConfigSecret` variables to the values file for the agent
- `kanikoPvcName` can be set to a pvc and the agent will then use that pvc to pass build contexts to kaniko builds. The pvc will be mounted in the agent pod and in kaniko pods.
- `kanikoDockerConfigSecret` can be set to the name of a `kubernetes.io/dockerconfigjson` secret. The agent will mount `.dockerconfigjson` from this secret into `/kaniko/.docker/config.json` in the kaniko pod.


charts/launch-agent/Chart.yaml
```diff
@@ -3,7 +3,7 @@ name: launch-agent
 icon: https://em-content.zobj.net/thumbs/240/apple/354/rocket_1f680.png
 description: A Helm chart for running the W&B Launch Agent in Kubernetes
 type: application
-version: 0.12.2
+version: 0.12.3
```
wandb-zacharyblasczyk, line 6: This seems like more of a feature than a patch right?

wandb-zacharyblasczyk (approved):

PR #85, opened 2024-02-29
chore: dont create managed cert in charts for google

adityachoudhari26:


charts/operator-wandb/templates/ingress.yaml
```diff
@@ -1,18 +1,5 @@
...
-  name: {{ .Release.Name }}-cert
-  namespace: {{ .Release.Namespace }}
-  labels:
-    {{- include "wandb.commonLabels" . | nindent 4 }}
-spec:
-  domains:
-  - {{ replace "https://" "" (replace "http://" "" .Values.global.host) }}
-{{- else }}
----
+{{- if ne .Values.ingress.issuer.provider "google" }}
```
jsbroks, line 2: `ne .Values.ingress.issuer.provider "google"` we can remove all these checks. If issuer is set to create, that use the logic, we don't need to check which type of provider is being set

jsbroks (approved):

jsbroks (approved):

PR #87, opened 2024-03-06
chore: Add nameOverride to ingress

adityachoudhari26:


jsbroks (changes requested):
helm automically handles namespace

nfoucha:
> helm automically handles namespace

Not what is trying to be solved with this PR.  That said, we should consider using a different key than `namespace` because of the relation to helm/k8s "namespace".  A customer looking at that value is going to be confused.

charts/operator-wandb/templates/ingress.yaml
```diff
@@ -24,7 +24,7 @@ spec:
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
-  name: {{ .Release.Name }}
+  name: {{ default .Release.Name .Values.ingress.name }}
```
nfoucha, line 27: Nit: The common pattern here is to use `nameOverride`

jsbroks (approved):

nfoucha (approved):

PR #91, opened 2024-03-18
feat(backend): Add flat runs field updater

nfoucha:
Adds a sub-chart for flat-runs-field-updater
Add kafka helpers
Remove value templating for producer/consumer urls.


charts/operator-wandb/charts/flat-runs-field-updater/templates/deployment.yaml
```diff
@@ -0,0 +1,155 @@
...
+            - name: GORILLA_RUN_UPDATE_SHADOW_QUEUE
+              value: >
+                {
+                  "overflow-bucket": {
+                    "store": "$(BUCKET)",
+                    "name": "wandb",
+                    "prefix": "wandb-overflow"
+                  },
+                  "subscriptions": {
+                    "flatRunFieldsUpdater": "kafka://$(KAFKA_CLIENT_USER):$(KAFKA_CLIENT_PASSWORD)@wandb-kafka:9092/flat-runs?consumer_group_id=deafult-group"
```
nfoucha, line 132: @jsbroks The idea here is that the topic, and potentially the consumer group as well, would be their own envvars/values.  We'll include the release name in the default value as we do elsewhere which should address your concern about namespacing the topic.

charts/operator-wandb/charts/app/templates/deployment.yaml
```diff
@@ -198,6 +198,31 @@ spec:
...
+              value: "{{ include "app.bucket" .}}"
+            - name: GORILLA_RUN_UPDATE_SHADOW_QUEUE
+              value: >
+                {
+                  "overflow-bucket": {
+                    "store": "$(OVERFLOW_BUCKET_ADDR)",
+                    "name": "wandb",
+                    "prefix": "wandb-overflow"
+                  },
+                  "foo": "bar",
```
jessicaxiang, line 223: was this left in by accident? 😂

charts/operator-wandb/charts/app/templates/deployment.yaml
```diff
@@ -198,6 +198,31 @@ spec:
...
+                  "addr": "kafka://$(KAFKA_CLIENT_USER):$(KAFKA_CLIENT_PASSWORD)@$(KAFKA_BROKER_HOST):$(KAFKA_BROKER_PORT)/flat-runs?producer_batch_bytes=1048576"
```
jessicaxiang, line 224: Is it too late to name this `run-updates-shadow` instead of `flat-runs` to match what we have for pubsub?

charts/operator-wandb/charts/flat-runs-field-updater/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+---
+apiVersion: v2
+name: flat-runs-field-updater
```
jessicaxiang, line 3: Could we name this `flat-run-fields-updater` (same with the name of this file) to match what we have in SaaS?

charts/operator-wandb/charts/flat-runs-field-updater/templates/deployment.yaml
jessicaxiang: Is it not too late to fix this typo? 😂 `deafult-group` -> `default-group`

charts/operator-wandb/charts/app/templates/deployment.yaml
jsbroks: Could we add namespacing

charts/operator-wandb/charts/flat-runs-field-updater/templates/hpa.yaml
```diff
@@ -0,0 +1,33 @@
...
+    {{- end }}
+    {{- if .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    - type: Resource
+      resource:
+        name: memory
+        target:
+          type: Utilization
+          averageUtilization: {{ .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    {{- end }}
+{{- end }}
```
jsbroks, line 33: Could we use the naming schema and patterns found in [bitnami](https://github.com/bitnami/charts/blob/5e4910e8bc2b7624e34f91950cd378bca420000b/bitnami/nginx-ingress-controller/templates/controller-hpa.yaml#L8)

charts/operator-wandb/charts/flat-run-fields-updater/templates/hpa.yaml
```diff
@@ -0,0 +1,33 @@
...
+          averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
+    {{- end }}
+    {{- if .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    - type: Resource
+      resource:
+        name: memory
+        target:
+          type: Utilization
+          averageUtilization: {{ .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    {{- end }}
```
jsbroks, line 32: Let's start by defining these statically for simplicity's sake. We can always add templating later if the need arises. This also prevents customers from accidentally modifying the values and it follows the pattern we are currently use in the codebase

charts/operator-wandb/charts/flat-run-fields-updater/templates/hpa.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if .Values.autoscaling.enabled }}
+---
+apiVersion: autoscaling/v2
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "flat-run-fields-updater.fullname" . }}
+  labels:
+    {{- include "flat-run-fields-updater.labels" . | nindent 4 }}
```
jsbroks, line 8:
```suggestion
  labels:
    {{- include "wandb.commonLabels" . | nindent 4 }}
    {{- include "flat-run-fields-updater.commonLabels" . | nindent 4 }}
    {{- include "flat-run-fields-updater.labels" . | nindent 4 }}
```

charts/operator-wandb/charts/flat-run-fields-updater/values.yaml
```diff
@@ -0,0 +1,77 @@
+---
+# Default values for flat-run-fields-updater.
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+enabled: false
```
jsbroks, line 5:
```suggestion
```

This value doesn't do anything. Its defined at the parent value.

charts/operator-wandb/charts/flat-run-fields-updater/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,13 @@
...
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "flat-run-fields-updater.serviceAccountName" . }}
+  labels:
+    {{- include "flat-run-fields-updater.labels" . | nindent 4 }}
+  {{- with .Values.serviceAccount.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
```
jsbroks, line 12: To improve code readability, let's apply the same labeling and annotation patterns across the codebase.
```suggestion
  labels:
    {{- include "wandb.commonLabels" . | nindent 4 }}
    {{- include "flat-run-fields-updater.commonLabels" . | nindent 4 }}
    {{- include "flat-run-fields-updater.labels" . | nindent 4 }}
    {{- if .Values.serviceAccount.labels -}}
    {{-   toYaml .Values.serviceAccount.labels | nindent 4 }}
    {{- end }}
  annotations:
    {{- if .Values.serviceAccount.annotations -}}
    {{-   toYaml .Values.serviceAccount.annotations | nindent 4 }}
    {{- end }}
```

charts/operator-wandb/charts/flat-run-fields-updater/templates/_helpers.tpl
```diff
@@ -0,0 +1,127 @@
...
+Returns a list of _pod_ labels to be shared across all
+flat-run-fields-updater deployments.
+*/}}
+{{- define "flat-run-fields-updater.podLabels" -}}
+{{- range $key, $value := .Values.pod.labels }}
+{{ $key }}: {{ $value | quote }}
+{{- end }}
+{{- end -}}
+
+
```
jsbroks, line 81:
```suggestion

```

charts/operator-wandb/templates/_kafka.tpl
```diff
@@ -51,4 +50,16 @@ Return kafka broker url port
...
+{{/*
+Return the kafka topic name for run-updates-shadow
+*/}}
+{{- define "wandb.kafka.runUpdatesShadowTopic" -}}
+{{- if eq .Values.global.kafka.runUpdatesShadowTopic "" -}}
+{{ printf "%s-%s" .Release.Name "run-updates-shadow" | trunc 63 | trimSuffix "-" }}
+{{- else -}}
+{{ .Values.global.kafka.runUpdatesShadowTopic }}
+{{- end -}}
+{{- end -}}
```
jsbroks, line 65: To keep things manageable, let's use predefined values instead customer specific ones. Customers likely won't interact with this directly, and it'll make our Helm chart easier to maintain. We can always add it later, but we cannot remove options later

```suggestion
{{/*
Return the kafka topic name for run-updates-shadow
*/}}
{{- define "wandb.kafka.runUpdatesShadowTopic" -}}
{{ printf "%s-%s" .Release.Name "run-updates-shadow" | trunc 63 | trimSuffix "-" }}
{{- end -}}
```

jsbroks (changes requested):

charts/operator-wandb/charts/flat-run-fields-updater/templates/hpa.yaml
nfoucha: I don't think statically defining these is a good approach for a couple of reasons.  First, this is the helm recommended file/format for an hpa and already has everything defined for us out of the box so we can hit the ground running when we're ready for hpa architecture.  Second, until we actually start using the system we're not going to have any real insight for what these values should be.  Leaving these as dynamic placeholders means we don't have to do the extra work of figuring out baselines before merging this in.

jsbroks (approved):

PR #95, opened 2024-03-20
chore: readiness and liveness probes for parquet

ramit-wandb:
I've thrown in what seemed like sensible defaults to me - please let me know if we want to edit these or template these out.


charts/operator-wandb/charts/parquet/templates/deployment.yaml
```diff
@@ -113,6 +113,26 @@ spec:
             {{- include "parquet.extraEnv" (dict "global" .Values.global "local" .Values) | nindent 12 }}
             {{- include "wandb.extraEnvFrom" (dict "root" $ "local" .) | nindent 12 }}
 
+          livenessProbe:
+            httpGet:
+              path: /ready 
+              port: 8087
+            initialDelaySeconds: 30
```
nfoucha, line 120: In future releases we may want to consider parameterizing these values so we can adjust on a per-install basis.  For now this is fine.

nfoucha (approved):

jsbroks (approved):

PR #100, opened 2024-03-26
chore: Email server

adityachoudhari26:


charts/operator-wandb/charts/app/templates/deployment.yaml
```diff
@@ -134,6 +134,10 @@ spec:
                   name: {{ include "app.fullname" . }}-config
                   key: SLACK_SECRET
                   optional: true
+            {{- if ne .Values.global.email.host "" }}
```
nfoucha, line 137: Since this is just related to the app, shouldn't the value be namespaced there instead of global?

jsbroks: Since we don't know if or when this will become its own service, or which services might need email functionality, it's safest to place it in the global configuration for now.

charts/operator-wandb/values.yaml
```diff
@@ -48,6 +48,12 @@ global:
   slack:
     secret: ""
     clientId: ""
+  
+  email:
```
jsbroks, line 52: maybe

```
email:
  stmp:
```

We might had support for other protocals

nfoucha: `pop`, `imap`, and `smtp` for completeness

jsbroks (approved):

PR #102, opened 2024-03-29
feat: Added internal ingress 

amanpruthi:


charts/operator-wandb/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.11.5
+version: 0.13.5
```
wandb-zacharyblasczyk, line 5: 0.12.0?

charts/operator-wandb/values.yaml
```diff
@@ -90,6 +90,22 @@ global:
...
+  tls: []
+  additionalHosts: []
+  class: ""
+  issuer:
+    create: false
+    provider: ""
+    server: https://acme-v02.api.letsencrypt.org/directory
+    email: support@wandb.com
+
+ingress_internal:
```
wandb-zacharyblasczyk, line 107: We don't use `_` I think camelCase is that is being standardized on. 

Also, what is the ing

wandb-zacharyblasczyk: Also, if this is for privatLink, maybe we should call this 
```
privatLinkIngress:
  create: false
...
```

or 
```
offlineIngress:
  create: false
...
```

We should probably document that this ingress is only needed for GCP/Azure/Bare metal?

wandb-zacharyblasczyk (commented):
@gls4 @nfoucha @jsbroks @adityachoudhari26

charts/operator-wandb/templates/ingress.yaml
```diff
@@ -80,3 +81,64 @@ spec:
               number: 8082
   {{- end }}
 {{- end }}
+{{- end }}
+---
+{{- if .Values.gcpPrivatLinkIngress.create }}
```
jsbroks, line 86:
```suggestion
{{- if .Values.ingress.secondary.create }}
```

Lets generalize and call it `secondary`

charts/operator-wandb/templates/ingress.yaml
```diff
@@ -80,3 +81,64 @@ spec:
...
+            port: 
+              number: 8080
+            {{- end }}
+      - pathType: Prefix
+        path: /console
+        backend:
+          service:
+            name: {{ $.Release.Name }}-console
+            port:
+              number: 8082
```
jsbroks, line 142: Could we move this to a tpl and import that here? This way we only need to modify it in one place

jsbroks (changes requested):

charts/operator-wandb/values.yaml
```diff
@@ -116,6 +117,21 @@ ingress:
     provider: ""
     server: https://acme-v02.api.letsencrypt.org/directory
     email: support@wandb.com
+  secondary:
+    create: true
```
jsbroks, line 121:
```suggestion
    create: false
```

jsbroks (approved):

jsbroks (approved):

charts/operator-wandb/templates/ingress.yaml
```diff
@@ -54,29 +55,54 @@ spec:
...
+  labels:
+    {{- include "wandb.commonLabels" . | nindent 4 }}
+    {{- if .Values.ingress.secondary.labels -}}
+    {{-   toYaml .Values.ingress.secondary.labels | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- if .Values.ingress.secondary.issuer.create }}
+    "cert-manager.io/issuer": "{{ .Release.Name }}-issuer"
+    "cert-manager.io/acme-challenge-type": "http01"
+    "kubernetes.io/ingress.allow-http" : "false"
```
zacharyblasczyk, line 82: Typo?

charts/operator-wandb/values.yaml
```diff
@@ -116,6 +117,21 @@ ingress:
     provider: ""
     server: https://acme-v02.api.letsencrypt.org/directory
     email: support@wandb.com
+  secondary:
+    create: false
+    install: true
```
zacharyblasczyk, line 122: Unused?

charts/operator-wandb/values.yaml
```diff
@@ -116,6 +117,21 @@ ingress:
...
+    annotations: {}
+    labels: {}
+    tls: []
+    additionalHosts: []
+    class: ""
+    issuer:
+      create: false
+      provider: ""
+      server: https://acme-v02.api.letsencrypt.org/directory
+      email: support@wandb.com
```
zacharyblasczyk, line 134: Do these need to be wired up?

zacharyblasczyk (commented):
Looks like a bit of copy paste.

PR #107, opened 2024-04-03
chore(backend): clean up runs v2 env variables

jessicaxiang:
Some light cleanup of variables that are not needed in certain deployments


yogeshg (approved):

charts/operator-wandb/charts/flat-run-fields-updater/templates/deployment.yaml
```diff
@@ -97,18 +97,6 @@ spec:
               value: "mysql://$(MYSQL_USER):$(MYSQL_PASSWORD)@$(MYSQL_HOST):$(MYSQL_PORT)/$(MYSQL_DATABASE)"
             - name: GORILLA_RUN_STORE
```
yogeshg, line 98: let's also include `GORILLA_RUN_STORE` with the `wandb_local_flat` in the release channels, so that this is enabled only for managed installs.

charts/operator-wandb/charts/flat-run-fields-updater/templates/deployment.yaml
```diff
@@ -97,18 +97,6 @@ spec:
...
               value: "mysql://$(MYSQL_USER):$(MYSQL_PASSWORD)@$(MYSQL_HOST):$(MYSQL_PORT)/wandb_local_flat?tls=preferred"
```
jsbroks, line 99:
```suggestion
```

jsbroks (approved):

PR #110, opened 2024-04-09
chore: Set gorilla oidc secret

adityachoudhari26:


nfoucha (approved):

wandb-zacharyblasczyk (approved):
I'm a bit worried that not everyone is setting `.Values.sso.clientSecret`, but better to make this change now rather than later.

venky-wandb:
Where did we land with merging this change? Was there any concern with adding this in?

dacbd:
appears to be superseded by changes in https://github.com/wandb/helm-charts/pull/260 or earlier

PR #112, opened 2024-04-12
fix: Banner templating

adityachoudhari26:


nfoucha (changes requested):
Those are rookie chart numbers, gotta bump those up

nfoucha (approved):

PR #113, opened 2024-04-13
fix(o11y): Enable JMX stats collection for Kafka

nfoucha:
Update kafka version to fix auth issues between metrics pod and broker pods
Enable prometheus endpoints for metrics and jmx stats
Add annotation for datadog autodiscovery of jmx stats


dacbd:
stale

PR #115, opened 2024-04-16
chore(weave): Use app fullname for weave auth

jwlee64:
weave was using host url for auth
this would fail if the url was ip allow listed
make the request to app.fullName(wandb-app) 

[dd error ](https://app.datadoghq.com/apm/trace/7825151400505903724?colorBy=service&env=managed-install&graphType=flamegraph&shouldShowLegend=true&sort=time&spanID=5329251616330542402&spanViewType=errors&timeHint=1713267779966.6892&view=spans)

screenshot showing the proper url in the env and that you can curl it (qa-aws)
Screenshot 2024-04-16 at 1 02 29 PM

### DIFF
--- a/charts/operator-wandb/charts/weave/templates/deployment.yaml
+++ b/charts/operator-wandb/charts/weave/templates/deployment.yaml
@@ -60,7 +60,7 @@ spec:
             - name: WEAVE_LOCAL_ARTIFACT_DIR
               value: /vol/weave/cache
             - name: WEAVE_AUTH_GRAPHQL_URL
-              value: {{ .Values.global.host }}/graphql
+              value: http://{{ include "weave.appFullname" . }}.default.svc.cluster.local:8080/graphql
             - name: WEAVE_SERVER_NUM_WORKERS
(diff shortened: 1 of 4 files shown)


charts/operator-wandb/charts/weave/templates/deployment.yaml
```diff
@@ -60,7 +60,7 @@ spec:
             - name: WEAVE_LOCAL_ARTIFACT_DIR
               value: /vol/weave/cache
             - name: WEAVE_AUTH_GRAPHQL_URL
-              value: {{ .Values.global.host }}/graphql
+              value: http://{{ include "weave.appFullname" . }}.default.svc.cluster.local:8080/graphql
```
nfoucha, line 63: The `default.svc.cluster.local` piece should be parameterized, at the very least the `default` part.  `.Release.Namespace` will probably give you what you want.

nfoucha (changes requested):

nfoucha (approved):

PR #127, opened 2024-05-16
✨ Review and Restrict Kubernetes Resource Permissions

velotioaastha:
1. The permissions of Kubernetes resources are minimized to what is essential for their operation, enhancing the cluster's security posture. Namespace-specific scoping will be applied where beneficial, providing an additional layer of containment and control.


charts/operator/templates/role.yaml
```diff
@@ -3,4 +3,83 @@ kind: ClusterRole
 metadata:
   name: {{ include "name" . }}-manager-role
 rules:
-{{ toYaml .Values.clusterRole.rules | indent 2 }}
\ No newline at end of file
+- apiGroups:
+  - '*'
```
jsbroks, line 7: I think this should be the default values for clusterRole.rules

zacharyblasczyk:
Yah, @danielpanzella went through and added a bunch of needed permissions already.

PR #131, opened 2024-05-21
chore(weave): Add weave private url to helm chart

jwlee64:
This var allows weave to use a more direct private url when making pod to pod request while maintaining the normal wandb base url for frontend requests
This fixes an issues where wandb pod ip allowlisting blocked the weave pod from making requests to the public ip

updates an import for a failing pod

### DIFF
--- a/charts/operator-wandb/charts/weave/templates/deployment.yaml
+++ b/charts/operator-wandb/charts/weave/templates/deployment.yaml
@@ -54,7 +54,7 @@ spec:
             - name: ONLY_SERVICE
               value: weave
             - name: WANDB_BASE_URL
-              value: {{ .Values.global.host }}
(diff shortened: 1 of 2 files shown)


charts/operator-wandb/charts/weave/templates/deployment.yaml
```diff
@@ -55,6 +55,8 @@ spec:
               value: weave
             - name: WANDB_BASE_URL
               value: {{ .Values.global.host }}
+            - name: WANDB_PRIVATE_BASE_URL
+              value: http://{{ include "weave.appFullname" . }}.{{ $.Release.Namespace }}.svc.{{ .Values.app.clusterDomain }}:8080
```
jwlee64, line 59: got some comments from nigel, updating and testing pr now

jsbroks, line 59:
```suggestion
              value: http://{{ include "weave.appFullname" . }}:8080
```

Why cannot we keep this simple?

PR #132, opened 2024-05-22
feat: Option to create cert for gce

adityachoudhari26:
should be a pure revert of this: https://github.com/wandb/helm-charts/commit/e5c061f07a3257ff3c120a8e3e70c322a8a44164

### DIFF
--- a/charts/operator-wandb/templates/ingress.yaml
+++ b/charts/operator-wandb/templates/ingress.yaml
@@ -5,7 +5,6 @@ apiVersion: networking.gke.io/v1
 kind: ManagedCertificate
 metadata:
   name: {{ .Release.Name }}-cert
-  namespace: {{ .Release.Namespace }}
   labels:
     {{- include "wandb.commonLabels" . | nindent 4 }}
 spec:
--- a/charts/operator-wandb/templates/ingress.yaml
+++ b/charts/operator-wandb/templates/ingress.yaml
@@ -1,4 +1,18 @@
 {{- if .Values.ingress.issuer.create }}
+{{- if eq .Values.ingress.issuer.provider "google" }}
+---
+apiVersion: networking.gke.io/v1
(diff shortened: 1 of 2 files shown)


charts/operator-wandb/templates/ingress.yaml
```diff
@@ -1,4 +1,18 @@
 {{- if .Values.ingress.issuer.create }}
+{{- if eq .Values.ingress.issuer.provider "google" }}
+---
+apiVersion: networking.gke.io/v1
+kind: ManagedCertificate
+metadata:
+  name: {{ .Release.Name }}-cert
+  namespace: {{ .Release.Namespace }}
```
jsbroks, line 8:
```suggestion
```

charts/operator-wandb/templates/ingress.yaml
```diff
@@ -1,4 +1,18 @@
...
+  labels:
+    {{- include "wandb.commonLabels" . | nindent 4 }}
+spec:
+  domains:
+  - {{ replace "https://" "" (replace "http://" "" .Values.global.host) }}
```
jsbroks, line 13:
```suggestion
  - $defaultHost
```

What if there are multiple hosts? I think we support extra hosts

charts/operator-wandb/templates/ingress.yaml
```diff
@@ -47,9 +66,11 @@ spec:
   {{- $defaultHost := include "defaultHost" . -}}
 
   {{- if .Values.ingress.issuer.create }}
+  {{- if ne .Values.ingress.issuer.provider "google" }}
   tls:
     - host: [{{ $defaultHost }}]
```
jsbroks, line 71: What if there are multiple hosts? I think we support extra hosts

jsbroks (approved):

PR #139, opened 2024-05-29
feat: Added azure sql and redis metrics support in otel agent

amanpruthi:


charts/operator-wandb/charts/otel/charts/agent/templates/_receivers.tpl
```diff
@@ -161,6 +161,22 @@ receivers:
...
+    resource_groups:
+      {{- range .Values.presets.receivers.azuremonitor.resource_groups }}
+      - {{ . }}
+      {{- end }}
+    auth: "workload_identity"
+    tenant_id: "${env:AZURE_TENANT_ID}"
+    client_id: "${env:AZURE_CLIENT_ID}"
+    federated_token_file: "${env:AZURE_FEDERATED_TOKEN_FILE}"
+    services: ["Microsoft.DBforMySQL/flexibleServers","Microsoft.Cache/Redis"]
+{{- end }}
```
jsbroks, line 178: Won't we want to only set these in azure terraform enviroments? Otherwise this would fail if its deployed in AWS or GCP

amanpruthi:
No need to update in helm

PR #145, opened 2024-06-10
fix: Templatize http proxy container

adityachoudhari26:


charts/operator/values.yaml
```diff
@@ -33,4 +33,8 @@ clusterRole:
       resources: ["*"]
       verbs: ["*"]
 
+httpProxy:
+  name: kube-rbac-proxy
+  image: gcr.io/kubebuilder/kube-rbac-proxy:v0.5.0
```
jsbroks, line 38:
```suggestion
  image:
    repository: gcr.io/kubebuilder/kube-rbac-proxy
    tag: v0.5.0
```

We dont need to template name

jsbroks (approved):

PR #146, opened 2024-06-11
fix: added missing mysql port variable to init container

mkaesz-wandb:
The wandb-app init container always uses MySQL default port for connection testing.  It's hanging in init forever when a non-standard MYSQL port is configured.


flamarion (changes requested):
You may want to [bump this](https://github.com/wandb/helm-charts/blob/main/charts/operator-wandb/Chart.yaml#L5) version to make the GHA check goes through.

flamarion (approved):
LGTM.

jsbroks (approved):

PR #149, opened 2024-06-18
chore: Default mysql to false

adityachoudhari26:


mkaesz-wandb:
@adityachoudhari26 Can we leave Redis by default on "true"?

jsbroks (approved):

PR #157, opened 2024-07-02
feat(weave): Add weave-trace

nickpenaranda:
- [x] Add weave trace server, with its REST API exposed at <host>/traces


charts/operator-wandb/charts/app/templates/deployment.yaml
```diff
@@ -114,6 +114,10 @@ spec:
               value: "http://{{ .Release.Name }}-parquet:8087"
             - name: PARQUET_ENABLED
               value: "true"
+            {{- if .Values.global.weaveTraces.enabled}}
```
nickpenaranda, line 117: to tell the frontend that traces are enabled

charts/operator-wandb/charts/weave-trace/templates/migrate-hook.yaml
```diff
@@ -0,0 +1,68 @@
+{{ if .Values.migration.useHook }}
```
nickpenaranda, line 1: We may be able to remove this.

The problem with this approach is that it doesn't work as a pre-install hook because it depends on the `Secret` containing the clickhouse credentials, and those don't yet exist when the job container starts.  It works fine if the `Secret` is provisioned ahead of time.  The benefit to this approach is it is technically more correct since we only want a single migration container running.

The alternative, the `initContainer` associated with the `weave-trace` pod, runs an init container for every pod.  This is mostly OK since the migration script quickly NOOPs when there is no need to migrate.

charts/operator-wandb/charts/weave/templates/deployment.yaml
```diff
@@ -53,16 +53,14 @@ spec:
...
               value: weave
-            - name: WANDB_BASE_URL
-              value: http://{{ include "weave.appFullname" . }}:8080
             - name: WANDB_PUBLIC_BASE_URL
               value: {{ .Values.global.host }}
             - name: WEAVE_LOG_FORMAT
               value: json
             - name: WEAVE_LOCAL_ARTIFACT_DIR
               value: /vol/weave/cache
-            - name: WEAVE_AUTH_GRAPHQL_URL
```
nickpenaranda, line 64: Fixing this while we're here.  This env var was deprecated a long time ago and this was not really doing anything.  We get the desired effect by setting `WANDB_BASE_URL`.  This may not be correct for all k8s providers

charts/operator-wandb/charts/weave-trace/templates/deployment.yaml
```diff
@@ -0,0 +1,168 @@
...
+              protocol: TCP
+          env:
+            - name: PORT
+              value: "8080"
+            - name: API_PATH_PREFIX
+              value: "/traces"
+            - name: WANDB_PUBLIC_BASE_URL
+              value: {{ .Values.global.host }}
+            - name: WANDB_BASE_URL
+              value: http://{{ include "weaveTrace.appFullname" . }}.{{ $.Release.Namespace }}.svc.{{ .Values.app.clusterDomain }}:8080/
```
jsbroks, line 82: Why do we need to set this? Should the service name be enough? We are only setting the service name in other cases so asking for consistency

charts/operator-wandb/charts/weave-trace/templates/migrate-hook.yaml
jsbroks: Can you also add the creation of the secret to the hook at the root level? Likewise we could add the creation of those secrets in this sub chart

charts/operator-wandb/charts/weave-trace/templates/deployment.yaml
nickpenaranda: `weave-trace` uses gorilla for auth.  we want to avoid hitting external dns and the additional hop routing through the ALB

charts/operator-wandb/charts/weave-trace/values.yaml
```diff
@@ -0,0 +1,60 @@
+nameOverride: ""
+fullnameOverride: ""
+
+# These need to be set as well if app name overrides are set
+app:
+  serviceName: "app"
+  clusterDomain: "cluster.local"
+  nameOverride: ""
+  fullnameOverride: ""
```
jsbroks, line 9: could you explain what this does and why we need it?

nickpenaranda: We need to resolve the cluster-local hostname for app.  We do something similar in `weave` https://github.com/wandb/helm-charts/blob/6a8df136066d848568b5e3bc7e73ffcc9b01bf21/charts/operator-wandb/charts/weave/values.yaml#L4

nickpenaranda: re: why, `weave-trace` uses gorilla for auth.  we want to avoid hitting external dns and the additional hop routing through the ALB.

charts/operator-wandb/charts/weave-trace/templates/deployment.yaml
jsbroks: sorry I guess my question su more ` http://{{ include "weaveTrace.appFullname" . }}:8080` as thats how we do it elsewhere in the code

charts/operator-wandb/charts/weave-trace/values.yaml
jsbroks: If we do `http://{{ include "weaveTrace.appFullname" . }}:8080` doesn't that resolve needing this section?

charts/operator-wandb/charts/weave-trace/templates/deployment.yaml
nickpenaranda: just tested this out, TIL! i will make this change

charts/operator-wandb/charts/weave-trace/values.yaml
nickpenaranda: yes, confirmed this works. thanks!

nickpenaranda: wait, `weaveTrace.appFullname` itself depends on `app.*`, to get rid of those `app` fields we'd have to do promote the app chart's name-related values to globals, right?

jsbroks: We do it like this: https://github.com/wandb/helm-charts/blob/6a8df136066d848568b5e3bc7e73ffcc9b01bf21/charts/operator-wandb/charts/console/templates/deployment.yaml#L57

While not ideal, it simplifies

jsbroks: we hardcode the service name

charts/operator-wandb/charts/weave-trace/templates/deployment.yaml
nickpenaranda: this is done

charts/operator-wandb/charts/weave-trace/values.yaml
nickpenaranda: now gone

jsbroks (approved):

PR #166, opened 2024-07-10
metadataStoreOverflow

jsbroks:


charts/operator-wandb/values.yaml
```diff
@@ -102,6 +102,13 @@ global:
     brokerPort: 9092
     runUpdatesShadowTopic: ""
 
+  metadataStoreOverflow:
+    enabled: false
+    bucket: "$(BUCKET)"
```
jsbroks, line 107:
```suggestion
```

charts/operator-wandb/values.yaml
```diff
@@ -102,6 +102,12 @@ global:
     brokerPort: 9092
     runUpdatesShadowTopic: ""
 
+  metadataStoreOverflow:
+    enabled: false
```
annirudh, line 106: So the idea is that we'll override this with channels?

annirudh (approved):

charts/operator-wandb/values.yaml
jsbroks: Yes

PR #173, opened 2024-07-11
feat(backend): enable subpath in GCP

levinandrew:
This PR enables support for subpaths in GCP https://github.com/wandb/terraform-google-wandb/pull/145/files

Testing:
Manually deployed with https://github.com/wandb/helm-charts/blob/levinandrew/byob-subpath/charts/operator-wandb/local-development.md
- when the new `bucket_path` in the above terraform, was NOT specified, when I ssh'd to the wandb-app pod, I saw 
```
wandb@wandb-app-65d8cb96-m7jb5:~$ env | grep -i bucket
BUCKET=gs://andrew-levin-gcp-intense-goblin
```
- when the new `bucket_path` was specified as `"nested/path/test", I saw
```
wandb@wandb-app-7cbb475896-79266:~$ env | grep -i nested
BUCKET=gs://andrew-levin-gcp-intense-goblin/nested/path/test
    "store": "gs://andrew-levin-gcp-intense-goblin/nested/path/test",
OVERFLOW_BUCKET_ADDR=gs://andrew-levin-gcp-intense-goblin/nested/path/test
GORILLA_FILE_STORE=gs://andrew-levin-gcp-intense-goblin/nested/path/test
```

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.14.16
+version: 0.14.17
 appVersion: 1.0.0
 icon: https://wandb.ai/logo.svg
 
--- a/charts/operator-wandb/charts/app/templates/_helpers.tpl
+++ b/charts/operator-wandb/charts/app/templates/_helpers.tpl
@@ -116,7 +116,7 @@ app deployments.
 {{- $bucket = printf "az://%s/%s" .Values.global.bucket.name .Values.global.bucket.path -}}
 {{- end -}}
 {{- if eq .Values.global.bucket.provider "gcs" -}}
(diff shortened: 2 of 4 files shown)


vanpelt (approved):

jsbroks (changes requested):
You cannot change this, as it will break all azure instances

jsbroks:
You have to keep the name as path

levinandrew:
> You have to keep the name as path

@jsbroks instead of renaming `path`, what about just adding `subpath` as a new field? 

then in tf azure we could do https://github.com/wandb/terraform-azurerm-wandb/compare/levinandrew/byob-subpath?expand=1
image

and in gcp/aws tf (where we don't use the `path` field yet) we could do https://github.com/wandb/terraform-google-wandb/pull/145/files
image

jsbroks:
I think we should leave it as path, since that is how its already templated and generally in infra, path references to what we are internally calling a subpath

zacharyblasczyk (approved):

jsbroks (approved):

PR #178, opened 2024-07-18
fix: Normalize the service account create logic for all wandb applications

danielpanzella:
Normalize the service account create logic for all wandb applications

https://github.com/wandb/terraform-google-wandb/pull/151

To support changes in terraform-google-wandb version 5.0 with `enable_workload_identity`


jsbroks (approved):
amazing ty

PR #187, opened 2024-07-25
fix: Correctly set up ddtrace via env vars for weave services

nickpenaranda:


charts/operator-wandb/charts/weave/templates/deployment.yaml
```diff
@@ -63,6 +63,12 @@ spec:
               value: {{ .Values.global.host }}
             - name: WEAVE_SERVER_NUM_WORKERS
               value: "4"
+            - name: DD_SERVICE
+              value: weave-python
```
zacharyblasczyk, line 67: I think you want this wrapped in a 
```helm
{{- if .Values.datadog.enabled }}
{{- end }}
```
Otherwise you will always be setting this.

zacharyblasczyk (changes requested):

charts/operator-wandb/charts/weave/templates/deployment.yaml
nickpenaranda: done. had to add the value too

zacharyblasczyk (approved):
🚢

zacharyblasczyk (approved):

PR #191, opened 2024-07-31
fix: Make console fully functional in non-restricted environments

danielpanzella:
- App doesn't need permissions
- Console needs access to operator logs for full functionality.  
  - We may want to create a cleaner solution for this later, but for now lets not break existing deployments


adityachoudhari26 (approved):

zacharyblasczyk (commented):
Looks Amazing!

PR #199, opened 2024-08-02
feat: Differentiate between bucket settings provided as default and set explicitly by the user for wandb managed installs

danielpanzella:
There are some scenarios where we want to provide setting for the storage bucket that are overwritten via console ui and this change should allow the console to know the source of the configuration


zacharyblasczyk (approved):
Tested and Approved.

jsbroks (approved):

PR #203, opened 2024-08-07
chore: add more granular extraEnvs for parquet

ramit-wandb:
Adds 2 extra "extraEnv" definitions:

1. `parquet.deployment.extraEnv` : Responsible for injecting environment variables into the parquet deployment *without* injecting them into the backfiller cronjob.
2. `parquet.cronjob.extraEnv` : Responsible for injecting environment variables into the parquet backfiller cronjob *without* injecting them into the deployment.

We still keep `parquet.extraEnv` around which is responsible for injecting env vars into both the deployment and the cronjob.

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.17.3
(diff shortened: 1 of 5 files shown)


dacbd:
@ramit-wandb, it looks like this got a bit abandoned, 😢 do you feel there is still value from this?

dacbd:
I'm going to close this, but we can always re-open and push it through if there is still value.

PR #213, opened 2024-09-05
chore: Enable EXPORTHISTORYTOPARQUET by default across all envs at 1min interval

ramit-wandb:


dacbd:
Stale, @ramit-wandb if you think this change should still be made, let us know.

ramit-wandb:
I think this is being superceded by the tshirt based deployment configuration. We do not need this PR.

PR #214, opened 2024-09-09
fix: Fix extraEnv reference

flamarion:
fix: Fix extraEnv reference

### DIFF
--- a/charts/operator/Chart.yaml
+++ b/charts/operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator
 description: A Helm chart for Weights & Biases operator
 type: application
-version: 1.3.0
+version: 1.3.1
 appVersion: "1.0.0"
 maintainers:
   - name: wandb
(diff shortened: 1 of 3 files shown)


zacharyblasczyk:
https://github.com/search?q=repo%3Awandb%2Fhelm-charts%20extraEnv&type=code 

@jsbroks, did we intentionally name these differently?

jsbroks (changes requested):
This is a breaking change, we cannot do this

flamarion:
Without this, we can't pass, for example, `HTTP_PROXY` env var to the Operator to pull the definitions and charts.
Is there any alternative?

zacharyblasczyk:
@flamarion, can you update all the refs to `extaEnv`.

flamarion:
Done!  Could you check if this is what you wanted, @zacharyblasczyk ?

flamarion:
@jsbroks  `{{- if .Values.extraEnv }}` was evaluated, but the `{{- toYaml .Values.extraEnvs | nindent 8 }}` never happened, so I don't think this would break stuff.

danielpanzella (approved):

jsbroks (approved):

PR #219, opened 2024-09-17
feat: Disable permissions for node-level metrics and logs

amanpruthi:


charts/operator/values.yaml
```diff
@@ -31,11 +31,6 @@ role:
       resources:
         - endpoints
         - ingresses
-        - nodes
-        - nodes/metrics
-        - nodes/spec
-        - nodes/stats
-        - nodes/proxy
```
jsbroks, line 38: I acutally think this will cause an issue, because causes lists out node infor, @danielpanzella we can drop support for this but wanted to raise it

jsbroks (approved):

danielpanzella (approved):

PR #221, opened 2024-09-18
feat: Init HPA workaround

zacharyblasczyk:
# HPA workaround

**Case A**: Default case. `.Values.glueSingleton.enabled = false` then

```yaml
app:
  glueSingleton:
    enabled: false
  autoscaling: 
    hpa:
      enabled: false
      maxReplicas: 10
      minReplicas: 5
  image:
    repository: wandb/local-dev
    tag: 0.60.0-pre-hpa.0
```
1. The app deployment is created with `GLUE_ENABLED = true` which is a *no-op*
2. The app pod will **not** set `ONLY_SERVICE`
3. The glue deployment not created

**This is a noop** for existing deployments. 

---

**Case B**: This is the case where we want to run the glue container in addition  to the app container. `.Values.glueSingleton.enabled = true` then:

```yaml
app:
  glueSingleton:
    enabled: true
  autoscaling:
    hpa:
      enabled: true
      maxReplicas: 10
      minReplicas: 5
  image:
    repository: wandb/local-dev
    tag: 0.60.0-pre-hpa.0
```

1. The app deployment is created with `GLUE_ENABLED = false`
2. The app pod will **not** set `ONLY_SERVICE`
3. The glue deployment is created with `GLUE_ENABLED = true`
4. The glue pod **will** set `ONLY_SERVICE = "gorilla-glue"`

# HPA note
HPA will only be enabled if both of these are set to true. So it is possible to run glue separately from the app pod and both of them be singletons. The glue pod will always be a singleton.

```yaml
app:
  glueSingleton:
    enabled: false
  autoscaling:
    hpa:
      enabled: false
```

# Final notes

To make this work we will be dependent on the next release of the wandb/local image. 
There is a pre-image that can be used for testing as shown above.

### DIFF
--- a/charts/operator-wandb/charts/app/templates/_deployment.tpl
+++ b/charts/operator-wandb/charts/app/templates/_deployment.tpl
@@ -20,7 +20,7 @@ metadata:
     {{- end }}
 spec:
   replicas: 1
-  {{- if and .Values.hpa.enabled .Values.glueSingleton.enabled }}
+  {{- if and .Values.autoscaling.hpa.enabled .Values.glueSingleton.enabled }}
   strategy:
     type: RollingUpdate
     rollingUpdate:
--- a/charts/operator-wandb/charts/app/templates/_deployment.tpl
+++ b/charts/operator-wandb/charts/app/templates/_deployment.tpl
@@ -1,0 +1,291 @@
+{{/*
+This template is used to generate the deployment for the app, and is used for both the non-glue and glue deployments.
+*/}}
+{{- define "app.deployment" -}}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "app.fullname" . }}{{ .suffix }}
+  labels:
+    {{- include "wandb.commonLabels" . | nindent 4 }}
+    {{- include "app.commonLabels" . | nindent 4 }}
+    {{- include "app.labels" . | nindent 4 }}
+    {{- if .Values.deployment.labels }}
+    {{-   toYaml .Values.deployment.labels | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- include "wandb.deploymentAnnotations" . | nindent 4 }}
+    {{- if .Values.deployment.annotations }}
+    {{-   toYaml .Values.deployment.annotations | nindent 4 }}
+    {{- end }}
+spec:
+  replicas: 1
+  {{- if .Values.hpa.enabled }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 0
+      maxSurge: 1
+  {{- end }}
+  selector:
+    matchLabels:
+      {{- include "wandb.selectorLabels" . | nindent 6 }}
+      {{- include "app.labels" . | nindent 6 }}
+  template:
+    metadata:
+      labels:
+        {{- include "wandb.podLabels" . | nindent 8 }}
+        {{- include "app.commonLabels" . | nindent 8 }}
+        {{- include "app.podLabels" . | nindent 8 }}
+        {{- include "app.labels" . | nindent 8 }}
+      annotations:
+        checksum/secret: {{ include (print $.Template.BasePath "/secrets.yaml") . | sha256sum }}
+        {{- if .Values.pod.annotations }}
+        {{-   toYaml .Values.pod.annotations | nindent 8 }}
+        {{- end }}
+    spec:
+      serviceAccountName: {{ include "app.serviceAccountName" . }}
+      {{- if .tolerations }}
+      tolerations:
+        {{- toYaml .tolerations | nindent 8 }}
+      {{- end }}
+      {{- include "wandb.nodeSelector" . | nindent 6 }}
+      {{- include "wandb.priorityClassName" . | nindent 6 }}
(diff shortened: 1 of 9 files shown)


zacharyblasczyk:
Looks like linting has been broken for a while.

zacharyblasczyk:
@jsbroks, do we need the kubernetes cluster for the wandb legacy chart?

mwhooker:
@zacharyblasczyk I'm gonna need a little bit of time to spin up on HPA to understand this PR fully. I'll be reading up on that today. If it's more urgent please ping me and I'll prioritize my +1

charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -0,0 +1,291 @@
...
+    {{- end }}
+spec:
+  replicas: 1
+  {{- if and .Values.hpa.enabled .Values.glueSingleton.enabled }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 0
+      maxSurge: 1
+  {{- end }}
```
zacharyblasczyk, line 29: Only setting the rollout strategt if both hpa and glueSingleton enabled.

charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -0,0 +1,291 @@
...
+            - name: gorilla-statsd
+              containerPort: 8125
+              protocol: TCP
+          env:
+            - name: GLUE_ENABLED
+              value: "{{ .glueSingletonEnabled }}"
+            {{- if .onlyService }}
+            - name: ONLY_SERVICE
+              value: {{ .onlyService }}
+            {{- end }}
```
zacharyblasczyk, line 104: Only service will be set if the glue deployment is enabled. 

Noop on the wandb pod for `GLUE_ENABLED` unless `glueSingletonEnabled` is also false from the template.

charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -0,0 +1,291 @@
...
+              path: /ready
+              port: http
+            initialDelaySeconds: 20
+            periodSeconds: 5
+            failureThreshold: 120
+          lifecycle:
+            preStop:
+              exec:
+                command: ["sleep", "25"]
+          {{- end }}
```
zacharyblasczyk, line 274: Only enabled for app deployment.

danielpanzella (approved):

charts/operator-wandb/charts/app/templates/_deployment.tpl
jsbroks: Why do we need to do all this? Why cannot we just write it out?

jsbroks: I would perfer to be explict then implicit. This adds some magic, that we plan on decoupling anyway

charts/operator-wandb/charts/app/templates/_helpers.tpl
```diff
@@ -49,8 +49,8 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 Selector labels
 */}}
 {{- define "app.selectorLabels" -}}
-app.kubernetes.io/name: {{ include "app.name" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
+app.kubernetes.io/name: {{ include "app.name" . }}{{ .suffix }}
+app.kubernetes.io/instance: {{ .Release.Name }}{{ .suffix }}
 {{- end }}
```
jsbroks, line 54: reason?

charts/operator-wandb/charts/app/values.yaml
```diff
@@ -9,6 +9,14 @@ image:
   pullPolicy: Always
   # pullSecrets: []
 
+glueSingleton:
+  enabled: false
+
+hpa:
+  enabled: false
+  minReplicas: 1
+  maxReplicas: 1
```
jsbroks, line 18: we don't need hpa, we just need to be able to set replica count. Since this is a hack we should just do replace count and do this after based on the design we go with?

charts/operator-wandb/charts/app/templates/_helpers.tpl
zacharyblasczyk: This is how it is determining the actual deployment, otherwise the clobber eachouther.

charts/operator-wandb/charts/app/templates/_deployment.tpl
zacharyblasczyk: This was added to make sure that when an upgrade to the spec happens, it actually rolls out properly one by one. 

Otherwise it will kill all the pods but one and replace them in bulk.

charts/operator-wandb/charts/app/values.yaml
zacharyblasczyk: Does that mean I should just completely remove `charts/operator-wandb/charts/app/templates/hpa.yaml`? 

Otherwise it will cause issues or it isn't used.

zacharyblasczyk: I think it will be more work to shift in the future tbh.

PR #222, opened 2024-09-20
feat: Support to pull MySQL credentials from secrets

amanpruthi:
```
Added support for mysql password and root password.
mysql:
  passwordSecret:
    name: ""
    rootKey: "MYSQL_ROOT_PASSWORD"
    userKey: "MYSQL_PASSWORD"
        
```


charts/operator-wandb/charts/mysql/templates/statefulset.yaml
```diff
@@ -49,12 +49,12 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "wandb.mysql.passwordSecret" . }}
-                  key: MYSQL_ROOT_PASSWORD
+                  key: {{ if .Values.global.mysql.passwordSecret.key }}{{ .Values.global.mysql.passwordSecret.key }}{{ else }}MYSQL_ROOT_PASSWORD{{ end }}
```
vanpelt, line 52: Is this right?  I imagine we would not want the `MYSQL_ROOT_PASSWORD` to be the same as the password that the application uses to access the database.  I think it should be a different key if we want to support someone setting it.

vanpelt (changes requested):

charts/operator-wandb/values.yaml
```diff
@@ -58,7 +58,10 @@ global:
     database: "wandb_local"
     user: "wandb"
     password: ""
-
+    passwordSecret:
+      name: ""
```
vanpelt, line 62: What is name?  Why is this here?

vanpelt: Ah, I see now.  `name` is the name of the Secret in the cluster, `key` is the identifier to grab within that secret.  We could do something like the following.  Right now it's all or nothing, if a user specifies a different secret they need to provide two passwords, otherwise the application password will have God access to the DB which is a big no no.

```
passwordSecret:
  name:
  rootKey: "MYSQL_ROOT_PASSWORD"
  userKey: "MYSQL_PASSWORD"

Comment (amanpruthi) [reply charts/operator-wandb/values.yaml:62]: Updated

File: charts/operator-wandb/charts/mysql/templates/statefulset.yaml [hunk 3]
```diff
@@ -49,12 +49,12 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "wandb.mysql.passwordSecret" . }}
-                  key: MYSQL_ROOT_PASSWORD
+                  key: {{ if .Values.global.mysql.passwordSecret.rootKey }}{{ .Values.global.mysql.passwordSecret.rootKey }}{{ else }}MYSQL_ROOT_PASSWORD{{ end }}
```
danielpanzella, line 52: Since we default `.Values.global.mysql.passwordSecret.rootKey` to `MYSQL_ROOT_PASSWORD` I'm not sure the conditional is doing anything here

charts/operator-wandb/templates/_mysql.tpl
```diff
@@ -1,9 +1,36 @@
...
+  {{- print .Release.Name "-mysql" -}}
 {{- end -}}
+{{- end -}}
+
+{{/*
+Return the key of the secret where the password is stored, considering if the custom key is defined
+*/}}
+{{- define "wandb.mysql.passwordSecret.key" -}}
+{{- if .Values.global.mysql.passwordSecret.userKey }}
+  {{- .Values.global.mysql.passwordSecret.userKey -}}
```
danielpanzella, line 17: I dont think this is neccessary since `.Values.global.mysql.passwordSecret.userKey` has a default value it will never be falsey, this define is probably redundant

charts/operator-wandb/templates/_mysql.tpl
```diff
@@ -1,9 +1,36 @@
...
+{{- else }}
+  MYSQL_PASSWORD
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the key of the secret where the root password is stored, considering if the custom root key is defined
+*/}}
+{{- define "wandb.mysql.passwordSecret.rootKey" -}}
```
danielpanzella, line 26: same as above

danielpanzella (changes requested):

danielpanzella (approved):

charts/operator-wandb/values.yaml
```diff
@@ -62,6 +62,10 @@ global:
     database: "wandb_local"
     user: "wandb"
     password: ""
+    passwordSecret:
+      name: ""
+      rootKey: "MYSQL_ROOT_PASSWORD"
```
danielpanzella, line 67: we don't actually seem to be using these, am I missing something?

danielpanzella (changes requested):

abhinavg6:
@vanpelt - Can you plz approve this? Waiting on the merge to allow customers to use this change. Thanks.

jsbroks (approved):

PR #223, opened 2024-09-23
feat: Support to pull Redis credentials from secrets in the CR

velotioaastha:
Added support to pull REDIS entire configuration from secrets from Secrets when deploying CRD from yaml file.
Secret should look like this :
```apiVersion: v1
kind: Secret
metadata:
  name: my-redis-secret
type: Opaque
data:
  REDIS_PASSWORD: bXlwYXNzd29yZA==   
  REDIS_CA_CERT: bXljYQ== 
  REDIS_PORT: NjM3OQ==
  REDIS_HOST: bXlyaWRpcy5leGFtcGxlLmNvbQ==
```

### DIFF
--- a/charts/operator-wandb/templates/_redis.tpl
+++ b/charts/operator-wandb/templates/_redis.tpl
@@ -1,33 +1,58 @@
 {{/*
-Return name of secret where redis information is stored
+Return name of secret where redis information is stored or fallback if not present
 */}}
 {{- define "wandb.redis.passwordSecret" -}}
-{{- print .Release.Name "-redis" -}}
+{{- if .Values.global.redis.secretName -}}
+  {{ .Values.global.redis.secretName }}
+{{- else -}}
+  {{- print .Release.Name "-redis" -}}
 {{- end -}}
+{{- end }}
 
 {{/*
 Return the redis port
 */}}
 {{- define "wandb.redis.port" -}}
-{{- print $.Values.global.redis.port -}}
+{{- $redisPort := .Values.global.redis.port -}}
+{{- if .Values.global.redis.secretName -}}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.global.redis.secretName -}}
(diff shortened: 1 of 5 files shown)


charts/operator-wandb/templates/_redis.tpl
```diff
@@ -1,33 +1,58 @@
...
+{{- end }}
 
 {{/*
 Return the redis port
 */}}
 {{- define "wandb.redis.port" -}}
-{{- print $.Values.global.redis.port -}}
+{{- $redisPort := .Values.global.redis.port -}}
+{{- if .Values.global.redis.secretName -}}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.global.redis.secretName -}}
```
danielpanzella, line 18: I dont think the port or host need to go in the secret, they should probably just be passed as normal values

charts/operator-wandb/templates/_redis.tpl
```diff
@@ -1,33 +1,58 @@
...
 
 {{/*
 Return the redis host
 */}}
 {{- define "wandb.redis.host" -}}
-{{- if eq .Values.global.redis.host "" -}}
-{{ printf "%s-%s" .Release.Name "redis-master" }}
+{{- $redisHost := .Values.global.redis.host -}}
+{{- if .Values.global.redis.secretName -}}
+  {{- $secret := lookup "v1" "Secret" .Release.Namespace .Values.global.redis.secretName -}}
```
danielpanzella, line 32: same as above

danielpanzella (changes requested):
I have some other thoughts, but I need to get some additional context from @jsbroks as to why we provide the redis creds as both a single connection string and multiple fields, and more imporant, what does the application actually use.

danielpanzella:
@velotioaastha The only way I could figure out how I wanted this implemented was to implement it myself. See https://github.com/wandb/helm-charts/pull/229

PR #229, opened 2024-10-03
feat: Support reading redis password from a user supplied secret

danielpanzella:
Reimplement https://github.com/wandb/helm-charts/pull/223, as the only way I could figure out exactly what I wanted was to do it.


zacharyblasczyk (approved):

mwhooker:
LGTM. Only question I have is if there's matching documentation to people know how to use this

PR #231, opened 2024-10-03
feat: Allow operator-wandb to accept license as a secret key value pair

velotioaastha:


mwhooker (approved):
lgtm

FYI we are going to be transitioning away from kube secrets. I won't ask that this be changed to come from google secrets because everything else in the file uses kube secrets, but it might be good to use it in the future

The only question I have is if we document these options anywhere

PR #238, opened 2024-10-09
fix: Removing enableAdminAPI from wandb chart's values.yaml

jms-wandb:
Removing the `enableAdminAPI` key from the wandb chart's values.yaml file. Even being set to `false` currently this has tripped up customer installations as they are tempted to set it to `true`. This has caused customers to be unable to login after deploying the application, either requiring manual intervention to correct it or a fresh start to redeploy.


dacbd:
stale

PR #242, opened 2024-10-16
Increase kafka replication factor

jessicaxiang:
Set replication factor to be 3 for kafka to have higher availability during leader elections. We have three brokers so 3 is the highest that we can go.


mwhooker (approved):
lgtm

PR #247, opened 2024-10-30
Support new dangerzone endpoint access from weave-trace server

KyleGoyette:
In:  https://github.com/wandb/core/pull/25229 we added a new endpoint, to be used internally to pass around sensitive information. This endpoint relies on authentication through JWTs, to allow other services in wandb to fetch customer secret values.

This PR:
1. Creates a new oidc reviewer cluster role binding that allows unauthenticated access to get the oidc configuration
2. Adds an environment variable `GORILLA_INTERNAL_JWT_SUBJECTS_TO_ISSUERS` to the app chart for gorilla to set up the issuer and subject relation.

This unlocks the weave playground which relies on fetching user secrets into the weave-trace service to make requests to llm services.

note: their is a PR in core (https://github.com/wandb/core/pull/25352) to make use of the environment variable structure used here.


charts/operator-wandb/charts/weave-trace/templates/_helpers.tpl
```diff
@@ -4,19 +4,19 @@
...
+{{- default "weave-trace" .Values.nameOverride | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
 {{/*
 Create a default fully qualified name for weave-trace. (Should be something like wandb-weave-trace)
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 If release name contains chart name it will be used as a full name.
 */}}
-{{- define "weaveTrace.fullname" -}}
+{{- define "weave-trace.weaveTrace.fullname" -}}
```
jsbroks, line 15: we shouldn't be including weave-traces infront of everything, thats inherited by helm

charts/operator-wandb/templates/oidc.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
jsbroks, line 1: This should all be properly templated out using chart prefix etc.

charts/operator-wandb/charts/weave-trace/templates/_helpers.tpl
```diff
@@ -4,19 +4,19 @@
...
 {{- if .Values.fullnameOverride }}
 {{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" }}
 {{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
+{{- $name := default "weave-trace" .Values.nameOverride }}
```
jsbroks, line 19: why are we hard coding this?

charts/operator-wandb/charts/weave-trace/templates/serviceaccount.yaml
```diff
@@ -1,17 +1,17 @@
...
+    {{- $weaveLabels := index .Values.global "weave-trace" "serviceAccount" "labels" }}
+    {{- if $weaveLabels -}}
+    {{-   toYaml $weaveLabels | nindent 4 }}
     {{- end }}
   annotations:
-    {{- if .Values.serviceAccount.annotations -}}
-    {{-   toYaml .Values.serviceAccount.annotations | nindent 4 }}
+    {{- $weaveAnnotations := index .Values.global "weave-trace" "serviceAccount" "annotations" }}
+    {{- if $weaveAnnotations -}}
+    {{-   toYaml $weaveAnnotations | nindent 4 }}
```
jsbroks, line 16: Why are we changing these patterns?

charts/operator-wandb/charts/weave-trace/templates/_helpers.tpl
KyleGoyette: This function is now called in the app server context so it was using the wrong name when called there. There's probably a better way for me to do it.

jsbroks: Whats do you mean by app server context?

danielpanzella: technically, you should not be calling a template across charts

charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -260,6 +260,10 @@ spec:
               value: {{ .Values.artifactsGc.NumWorkers | quote }}
             - name: GORILLA_ARTIFACTS_GC_DELETE_FILES_NUM_WORKERS
               value: {{ .Values.artifactsGc.DeleteFilesNumWorkers | quote }}
+            {{- if index .Values.global "weave-trace" "enabled" }}
+            - name: GORILLA_INTERNAL_JWT_SUBJECTS_TO_ISSUERS
+              value: '{"system:serviceaccount:default:{{ include "weave-trace.weaveTrace.serviceAccountName" . }}": "https://kubernetes.default.svc.cluster.local"}'
+            {{- end }}
```
jsbroks, line 266: Something seems wrong here. We also cannot do direct namespace references

danielpanzella (approved):

PR #248, opened 2024-10-31
chore(artifacts): expose artifacts garbage-collection env vars

ibindlish:
Follow up core PR: https://github.com/wandb/core/pull/25284

Expose artifact GC's glue task arguments as environment variables for on-prem instances so that its possible to configure them via the UI. 

Context: 
Meta's instance is facing issues due to the artifact gc glue task and they can't turn it off so we need to find a better way for them to configure the glue task's arguments


charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -254,6 +254,12 @@ spec:
                   },
                   "addr": "kafka://$(KAFKA_CLIENT_USER):$(KAFKA_CLIENT_PASSWORD)@$(KAFKA_BROKER_HOST):$(KAFKA_BROKER_PORT)/$(KAFKA_TOPIC_RUN_UPDATE_SHADOW_QUEUE)?producer_batch_bytes=1048576&num_partitions=$(KAFKA_RUN_UPDATE_SHADOW_QUEUE_NUM_PARTITIONS)&replication_factor=3"
                 }
+            - name: GORILLA_ARTIFACTS_GC_BATCH_SIZE
+              value: {{ .Values.global.artifactsGc.BatchSize }}
```
danielpanzella, line 258: Do these values need to be available globally or can we scope these to the application itself `{{ .Values.app.artifactsGc.BatchSize }}` and also move them in the values.yaml

danielpanzella (changes requested):

charts/operator-wandb/charts/app/templates/_deployment.tpl
danielpanzella: yeah, the difference is that if in .Values.app only the app chart (which is where glue runs) will see those values, but in global they are passed to all the charts.

ibindlish: Ah, i see. Makes sense, thank you!

ibindlish: I moved the values to the `app` values file instead: https://github.com/wandb/helm-charts/blob/main/charts/operator-wandb/charts/app/values.yaml

and updated the vars to reference {{ .Values.artifactsGc.BatchSize }} 

Is this right?

danielpanzella: Yes, that is correct, i had a mistake in my previous comment, but you got the reference correct

danielpanzella (approved):

PR #255, opened 2024-11-04
feat: Add support to pull CA Certs from configMap

velotioaastha:
Add support to pull the list of CA Certs from configMaps (W&B App)

Create configmap
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-ca-certs
data:
  ca-cert1.crt: |
    -----BEGIN CERTIFICATE-----
    ...
    -----END CERTIFICATE-----
  ca-cert2.crt: |
    -----BEGIN CERTIFICATE-----
    ...
    -----END CERTIFICATE-----
```
values.yaml 
configMapName: "custom-ca-certs"

### DIFF
--- a/charts/operator-wandb/charts/app/templates/_deployment.tpl
+++ b/charts/operator-wandb/charts/app/templates/_deployment.tpl
@@ -73,10 +73,15 @@ spec:
               mountPath: /etc/ssl/certs/redis_ca.pem
               subPath: redis_ca.pem
             {{- end }}
+            {{- if .Values.global.configMapName }}
+            - name: wandb-ca-certs
+              mountPath: /usr/local/share/ca-certificates/
+            {{- else if .Values.global.customCACerts }}
             {{- range $index, $v := .Values.global.customCACerts }}
             - name: wandb-ca-certs
               mountPath: /usr/local/share/ca-certificates/customCA{{$index}}.crt
               subPath: customCA{{$index}}.crt
+            {{- end }}
             {{- end }}
           ports:
             - name: http
@@ -300,7 +305,11 @@ spec:
               - key: REDIS_CA_CERT
                 path: redis_ca.pem
         {{- end }}
-        {{- if .Values.global.customCACerts }}
+        {{- if .Values.global.configMapName }}
+        - name: wandb-ca-certs
+          configMap:
+            name: {{ .Values.global.configMapName }}
+        {{- else if .Values.global.customCACerts }}
         - name: wandb-ca-certs
           configMap:
             name: {{ include "wandb.fullname" . }}-ca-certs
--- a/charts/operator-wandb/charts/app/templates/_deployment.tpl
+++ b/charts/operator-wandb/charts/app/templates/_deployment.tpl
@@ -73,10 +73,11 @@ spec:
               mountPath: /etc/ssl/certs/redis_ca.pem
               subPath: redis_ca.pem
             {{- end }}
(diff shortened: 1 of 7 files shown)


velotioaastha:
support for Parquet to pull ca certs from configmap can be added after this [PR](https://github.com/wandb/helm-charts/pull/254) gets merged

flamarion:
@velotioaastha, will the configMap with multiple keys with your proposed configuration become multiple certificate files in the `/usr/local/share/ca-certificates/`?
I didn't test it, but I wonder if simply mounting it as is will have the effect we expect: multiple cert files.
If this is the expected outcome, could you add a line explaining that CM key `MUST` be `something.CRT`?
We expect `crt` files to be parsed by the `update-ca-certificates` program used by Ubuntu to add the certs to the system ca store.

flamarion (commented):
@velotioaastha, will the configMap with multiple keys with your proposed configuration become multiple certificate files in the `/usr/local/share/ca-certificates/`?
I didn't test it, but I wonder if simply mounting it as is will have the effect we expect: multiple cert files.
If this is the expected outcome, could you add a line explaining that CM key `MUST` be `something.CRT`?
We expect `crt` files to be parsed by the `update-ca-certificates` program used by Ubuntu to add the certs to the system ca store.

velotioaastha:
Yes, with the configuration proposed earlier, mounting a ConfigMap with multiple keys will indeed result in multiple files within the specified directory (/usr/local/share/ca-certificates/), with each key in the ConfigMap corresponding to a separate file. However, for update-ca-certificates to recognize and add these certificates to the system CA store on Ubuntu, each file name in the directory must end with .crt (as given in the description)

Where do you propose I add this line?

flamarion:
Thank you for your reply @velotioaastha 
It would be great if you could add the guidance right above the configuration [Here](https://github.com/wandb/helm-charts/pull/255/files#diff-49582d692acfce2358472a39bb9148cf4e8f2c637f814b07665c7d3d8797933aR133) for example.
This is important to ensure customers create the `crt` files.

flamarion (approved):

charts/operator-wandb/charts/app/templates/configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.global.customCACerts }}
+{{- if and (not .Values.global.configMapName) .Values.global.customCACerts }}
```
danielpanzella, line 1: I think the chart should probably support both a user supplied config map and individual certs passed at the same time.  Theres no reason to force one or the other

charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -73,11 +73,16 @@ spec:
               mountPath: /etc/ssl/certs/redis_ca.pem
               subPath: redis_ca.pem
             {{- end }}
+            {{- if .Values.global.configMapName }}
+            - name: wandb-ca-certs
+              mountPath: /usr/local/share/ca-certificates/
+            {{- else if .Values.global.customCACerts }}
```
danielpanzella, line 79: I think we can allow both at the same time

danielpanzella (approved):

velotioaastha:
We have closed https://github.com/wandb/helm-charts/pull/254 and added the same changes here to support pull CA Certs from configmap for parquet as well
Please re-review @flamarion @danielpanzella

danielpanzella (approved):

PR #257, opened 2024-11-06
feat: Support secret keys configurable

amanpruthi:


flamarion (commented):
I think would be good to default all secrets to the previous key, like you did for Redis https://github.com/wandb/helm-charts/pull/257/files#diff-2820a1c4d6acfec8bdf8acd70e1a305cb1f536ae092834bd566632bcd86bbafdR134 
Another aspect I believe is necessary, is to make this optional.
The reason is because if some customer is already using this config and the operator pull the latest version of chart, this may break because the secret key will be required.

danielpanzella (changes requested):
As pointed out by @flamarion [here](https://github.com/wandb/helm-charts/pull/257#pullrequestreview-2417935414) we need the values for the key to default to the previous value.

velotioaastha:
> Another aspect I believe is necessary, is to make this optional.
> The reason is because if some customer is already using this config and the operator pull the latest version of chart, this may break because the secret key will be required.

hey @danielpanzella @flamarion , we've added this already, you can check this here - 

For bucket - 
https://github.com/wandb/helm-charts/pull/257/files#diff-46611996f4a6fc6e0ffb1724cbd796c7fd5e7008c5c956ec1c071b38abf58d7bR21

For mysql - 
https://github.com/wandb/helm-charts/pull/257/files#diff-ce26fb708141f65196c806cca1ffc3d75fc1cda8c0a7ff615d6a097ae24abcb1R19
https://github.com/wandb/helm-charts/pull/257/files#diff-ce26fb708141f65196c806cca1ffc3d75fc1cda8c0a7ff615d6a097ae24abcb1R30

For redis - 
https://github.com/wandb/helm-charts/pull/257/files#diff-2820a1c4d6acfec8bdf8acd70e1a305cb1f536ae092834bd566632bcd86bbafdR134

danielpanzella (approved):

PR #258, opened 2024-11-09
Remove default kubernetes api issuer for JWT map

KyleGoyette:
Before we defaulted to the k8s api open id configuration. But unfortunately, on EKS the jwks uri returned by the k8s api is an internal IP address that we can't connect to (see [here](https://github.com/aws/containers-roadmap/issues/2234)).

Instead we need to use the actual JWT issuer in the helm charts which differs between GKE, EKS and AKS.

This PR removes the k8s api endpoint as the default, and instead asks users to fill in the issuer. The environment variable is not set in the app pod if the issuer is left blank

### DIFF
--- a/charts/operator-wandb/charts/app/values.yaml
+++ b/charts/operator-wandb/charts/app/values.yaml
@@ -74,6 +74,11 @@ artifactsGc:
   NumWorkers: 0
   DeleteFilesNumWorkers: 0
 
+# internalJWTMap is used to map service accounts to their issuers.
+# The issuer should be the issuer URL based on your cloud environment:
+# For GCP: https://container.googleapis.com/v1/projects/<project-id>/locations/<location>/clusters/<cluster-name>
+# For AWS: https://oidc.eks.<region>.amazonaws.com/id/<cluster-id>
+# For Azure: https://sts.windows.net/<tenant-id>/
 internalJWTMap:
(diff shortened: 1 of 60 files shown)


charts/operator-wandb/charts/app/values.yaml
```diff
@@ -74,6 +74,7 @@ artifactsGc:
   NumWorkers: 0
   DeleteFilesNumWorkers: 0
 
+# internalJWTMap is used to map service accounts to their issuers.
```
KyleGoyette, line 77: Should I make a comment here to outline how people can fill this in depending on their cluster?

danielpanzella: Probably would help.  do we need to be setting this in our terraform?

KyleGoyette: We're setting it internally here: https://github.com/wandb/core/pull/25402/files#diff-768fd3d816a568ef4f97bd468a85b8b652eb4aa9b3343a813e511d514b5aa916R307-R310

KyleGoyette: Added a comment indicating the kubectl command you can use to get this for your cluster.

danielpanzella (approved):
This will require additional changes in the terraform to work in dedicated

PR #260, opened 2024-11-12
feat: Init glue

zacharyblasczyk:


danielpanzella (approved):

charts/operator-wandb/values.yaml
```diff
@@ -324,3 +328,93 @@ kafka:
...
+    GORILLA_CASBIN_ADDRESS:
+      value: '{{ include "wandb.mysql" . | trim }}'
+    GORILLA_RUN_STORE:
+      value: '{{ include "wandb.mysql" . | trim }}'
+    GORILLA_FILE_STREAM_STORE_ADDRESS:
+      value: '{{ include "wandb.mysql" . | trim }}'
+    GORILLA_HISTORY_STORE:
+      value: 'http://{{ .Release.Name }}-parquet:8087/_goRPC_,{{ include "wandb.mysql" . | trim }}'
+  envFrom:
+    "wandb-mysql": "secretRef"
```
zacharyblasczyk, line 372: Planning to change this at a later date.

annirudh (approved):
Looks great!

annirudh:
Reminder: When this lands or https://github.com/wandb/helm-charts/pull/264 lands, we'll need to make sure Glue wires up to executor.

zacharyblasczyk:
> Reminder: When this lands or #264 lands, we'll need to make sure Glue wires up to executor.

@annirudh , what needs to be done exactly to wire it up?

annirudh:
> > Reminder: When this lands or #264 lands, we'll need to make sure Glue wires up to executor.
> 
> @annirudh , what needs to be done exactly to wire it up?

Something like these lines: https://github.com/wandb/helm-charts/pull/264/files#diff-eaf259f2094a11f6c98bcde786fa801443d5aba03722a86a5be3c0ea3ebd1b6bR116

charts/operator-wandb/templates/gorilla.yaml
```diff
@@ -0,0 +1,93 @@
...
+  GORILLA_DEFAULT_REGION: "{{ default .Values.global.cloudProvider "minio" }}-{{ .Values.global.bucket.region | default .Values.global.defaultBucket.region }}"
+  GORILLA_FILE_STORE: "{{ include "wandb.bucket" . | trim }}"
+  GORILLA_STORAGE_BUCKET: "{{ include "wandb.bucket" . | trim }}"
+  {{- if .Values.global.executor.enabled }}
+  GORILLA_TASK_QUEUE: "{{ include "app.redis" . | trim }}"
+  GORILLA_TASK_QUEUE_WORKER_ENABLED: "false"
+  {{- else }}
+  GORILLA_TASK_QUEUE: "noop://"
+  GORILLA_TASK_QUEUE_WORKER_ENABLED: "false"
+  {{- end }}
```
zacharyblasczyk, line 93: @annirudh please take a look at this.

Also, I see this in the app pod: 
``` yaml
           - name: GORILLA_TASK_QUEUE
              value: "{{ include "app.redis" . | trim }}"
            - name: GORILLA_TASK_QUEUE_MONITOR_PORT
              value: "10000"
            - name: GORILLA_TASK_QUEUE_WORKER_ENABLED
              value: "false"
            - name: GORILLA_CLEAR_TASK_DEDUPE_KEY_ENABLED
              value: "false"
```

Do we need the other values always set as well?

annirudh: This is intended to be the defaults, right? these look right to me.

Tangential: on a deploy though, is the system smart enough to know that it needs to update the config maps before it can spin up new pods? What if a new pod starts up before the config map has been updated?

charts/operator-wandb/templates/mysql.yaml
```diff
@@ -16,4 +16,17 @@ data:
...
+kind: ConfigMap
+metadata:
+  name: {{ .Release.Name }}-mysql-configmap
+  labels:
+    {{- include "wandb.commonLabels" . | nindent 4 }}
+data:
+  MYSQL_PORT: "{{ include "wandb.mysql.port" . }}"
+  MYSQL_HOST: "{{ include "wandb.mysql.host" . }}"
+  MYSQL_DATABASE: "{{ include "wandb.mysql.database" . }}"
+  MYSQL_USER: "{{ include "wandb.mysql.user" . }}"
```
annirudh, line 32: Super duper nit and out-of-scope for this PR: We are inconsistent about ending files with newlines. Should we decide one way or the other and add a linter to enforce?

charts/operator-wandb/templates/gorilla.yaml
zacharyblasczyk: We will have a sha of the CM's and Secrets to force the update of the deployment. 

```yaml
metadata:
  annotations:
    checksum/config: "<sha_value>"
    checksum/secret: "<sha_value>"
```

charts/operator-wandb/templates/mysql.yaml
zacharyblasczyk: We can think about this. I can add an issue and we can pick that up at a later point.

zacharyblasczyk: https://github.com/wandb/helm-charts/issues/274 
Basic placeholder for now.

danielpanzella (approved):

PR #262, opened 2024-11-15
feat: WIP - Chart for filestream service

danielpanzella:
Chart for filestream service

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -223,6 +223,12 @@ flat-run-fields-updater:
     repository: wandb/local
     tag: latest
 
+filestream:
+  install: false
+  image:
+    repository: wandb/local-dev
+    tag: 0.63.0-danielpanzella-filestream-local.0
+
 mysql:
   install: false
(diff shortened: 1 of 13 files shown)


zacharyblasczyk (approved):

charts/operator-wandb/values.yaml
```diff
@@ -223,6 +226,12 @@ flat-run-fields-updater:
     repository: wandb/local
     tag: latest
 
+filestream:
+  install: false
+  image:
+    repository: wandb/local-dev
+    tag: 0.63.0-danielpanzella-filestream-local.1
```
annirudh, line 233: Is this intentionally checked in hardcoded?

danielpanzella: yeah, since that branch is the only one that works, we'll clean this up when everything settles on the core side

danielpanzella: it's overridable at several points.

annirudh: makes sense!

PR #265, opened 2024-11-20
🐛 fix bucket secret configuration

velotioaastha:


charts/operator-wandb/charts/app/templates/_helpers.tpl
```diff
@@ -123,11 +123,13 @@ app deployments.
...
-{{- $bucket = printf "s3://%s:%s@%s/%s" $bucketValues.accessKey $bucketValues.secretKey $bucketValues.name (default "" $bucketValues.path) -}}
-{{- else -}}
-{{- $bucket = printf "s3://%s/%s" $bucketValues.name (default "" $bucketValues.path) -}}
+{{- $credentials := "" -}}
+{{- if and (not $bucketValues.accessKey) (not $bucketValues.secretKey) $bucketValues.secretName -}}
+{{- $credentials = printf "$(ACCESS_KEY):$(SECRET_KEY)@" -}}
+{{- else if and $bucketValues.accessKey $bucketValues.secretKey -}}
+{{- $credentials = printf "%s:%s@" $bucketValues.accessKey $bucketValues.secretKey -}}
 {{- end -}}
+{{- $bucket = printf "s3://%s%s/%s" $credentials $bucketValues.name (default "" $bucketValues.path) -}}
```
flamarion, line 132: This is only required if the user configures the access and secret keys or the Secret name.
There's a possibility users rely on IAM, so they will not need to inform the access/secret pair or the secret name.
I think tweaking the ternary operator should get this done.

charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -208,6 +208,26 @@ spec:
...
+            - name: SECRET_KEY
+              {{- if .Values.global.bucket.secretKey }}
+              value: "{{ .Values.global.bucket.secretKey | default .Values.global.defaultBucket.secretKey }}"
+              {{- else }}
+              valueFrom:
+                secretKeyRef:
+                  name: "{{ include "wandb.bucket.secret" . }}"
+                  key: {{ .Values.global.bucket.secretAccessKeyName }}
+                  optional: true
+              {{- end }}
```
flamarion, line 230: This block can be optional and conditioned to the access/secret pair or secret name.
If the user doesn't pass any of those, we don't need to set these env vars.
The reason is that users can rely on IAM when using AWS.

flamarion (changes requested):
Please, check if the comments make sense and let me know if I missed something.

charts/operator-wandb/charts/app/templates/_helpers.tpl
velotioaastha: @flamarion 
If the user doesn’t pass any of those and is relying on IAM
In this case, should bucket secret be created or not?
If not - what output are we expecting in "AZURE_STORAGE_KEY" ?

flamarion: The `AZURE_STORAGE_KEY` env var is required only when the schema is `az://....` 
For `s3://`, the user can or cannot use the secret pair. since it can use IAM in case they are in AWS

flamarion (approved):
LGTM

PR #268, opened 2024-11-26
fix: Additional bigtable and pubsub fixes

danielpanzella:
- Allow bigtablev2 to be enabled via boolean `global.bigtable.enablev2`

### DIFF
--- a/charts/operator-wandb/charts/filestream/templates/_helpers.tpl
+++ b/charts/operator-wandb/charts/filestream/templates/_helpers.tpl
@@ -131,13 +131,17 @@ Create the name of the service account to use
 {{- define "filestream.fileStreamWorkerSource" -}}
 {{- if .Values.global.pubSub.enabled -}}
 pubsub:/{{ .Values.global.pubSub.project }}/{{ .Values.global.pubSub.filestreamTopic }}/{{ .Values.pubSub.subscription }}
-{{- else }}
-{{- end }}
-{{- end }}
+{{- else -}}
+{{- end -}}
+{{- end -}}
 
 {{- define "filestream.fileStreamWorkerStore" -}}
+{{- $fileStreamStore := "" -}}
 {{- if .Values.global.bigTable.enabled -}}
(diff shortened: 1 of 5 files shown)


charts/operator-wandb/charts/app/templates/_deployment.tpl
```diff
@@ -267,6 +267,10 @@ spec:
                   },
                   "addr": {{ include "app.runUpdateShadowTopic" . | quote }}
                 }
+            - name: GORILLA_SETTINGS_CACHE
+              value: "{{ include "executor.redis" . | trim }}"
```
annirudh, line 271: This needs t obe `app.redis`, right?

danielpanzella: yeah, these should be app

charts/operator-wandb/charts/filestream/templates/_helpers.tpl
```diff
@@ -131,13 +131,17 @@ Create the name of the service account to use
...
+{{- end -}}
 
 {{- define "filestream.fileStreamWorkerStore" -}}
-{{- if .Values.global.bigTable.enabled -}}
-bigtablev3://{{ .Values.global.bigTable.project }}/{{ .Values.global.bigTable.instance }},bigtablev2://{{ .Values.global.bigTable.project }}/{{ .Values.global.bigTable.instance }}
-{{- else }}
-{{- end }}
-{{- end }}
+{{- $fileStreamStore := "" -}}
+{{- if .Values.global.bigtable.v3.enabled -}}
```
zacharyblasczyk, line 140: Is this meant to be embedded?

zacharyblasczyk: Should it be possible to enable v2 without v3.

annirudh: Yes, it is possible to enable V2 without V3.

zacharyblasczyk: Okay, looks like this needs to be changed then.

annirudh: Let me fix this

annirudh: Fixed

PR #273, opened 2024-12-03
feat: Add support for topologySpreadConstraints

velotioaastha:
## Summary by CodeRabbit

- **New Features**
  - Introduced an optional configuration for enhanced pod distribution, allowing for improved high availability and fault tolerance across deployments. This setting is now available for multiple components but remains disabled by default.
  
- **Chores**
  - Updated the release version to 0.26.7.

### DIFF
--- a/charts/operator-wandb/Chart.lock
+++ b/charts/operator-wandb/Chart.lock
@@ -1,66 +1,66 @@
 dependencies:
-- name: app
-  repository: file://charts/app
-  version: 0.1.0
-- name: wandb-base
-  repository: file://../wandb-base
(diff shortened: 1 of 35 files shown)


danielpanzella (changes requested):
redis-exporter, mysql-exporter, stackdriver, and prometheus are going to be run as single instances so not sure adding topology constraints makes any sense.  I'm not sure we should be applying these to things like mysql or redis either.  so probably just dont include this for any of the non-application deployments

dacbd:
Stale, if this is still needed we can readdress it.

PR #283, opened 2024-12-13
feat: GCS FUSE support

annirudh:


charts/operator-wandb/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.21.6
+version: 0.21.7
```
danielpanzella, line 5: technically this should be 0.22.0

charts/operator-wandb/charts/parquet/templates/deployment.yaml
```diff
@@ -32,6 +33,16 @@ spec:
         {{- include "parquet.podLabels" . | nindent 8 }}
         {{- include "parquet.labels" . | nindent 8 }}
       annotations:
+        {{- if .Values.fuse.enabled }}
+        {{- if eq $bucketProvider "gcs" }}
```
danielpanzella, line 37: is this going to drop a sidecar into the pod?

danielpanzella: probably not, just worth knowing if we see issues with pod creation in the future.

zacharyblasczyk (approved):

PR #295, opened 2024-12-18
fix: Creating secret when secretName is provided

dacbd:
Prevents us from trying to create a redis secret of the same name as the users provided secret.

Other changes:
- template reference `wandb.redis.passwordSecret` changed to `wandb.redis.secretName` better reflects what the value is.
- moves `REDIS_CA_CERT` to the configmap

ON HOLD until after the holidays, need to cc @flamarion for validation.


dacbd:
No longer able to reproduce the orginal error, and other changes have been made to the redis secret since the original error was reported. [ref](https://weightsandbiases.slack.com/archives/C06RG0ERD0X/p1742597675758349?thread_ts=1734364729.418499&cid=C06RG0ERD0X)

PR #298, opened 2025-01-02
feat: Init Settings Migration Job

zacharyblasczyk:


charts/operator-wandb/templates/gorilla.yaml
```diff
@@ -10,10 +10,10 @@ stringData:
   {{- else }}
   GORILLA_EMAIL_SINK: "https://api.wandb.ai/email/dispatch"
   {{- end }}
-  SLACK_SECRET: {{ default "" .Values.global.slack.secret | b64enc }}
+  SLACK_SECRET: {{ default "" .Values.global.slack.secret }}
   {{- if and (not .Values.global.licenseSecret.name) (not .Values.global.licenseSecret.key) .Values.global.license }}
-  LICENSE: {{ .Values.global.license | b64enc }}
-  GORILLA_LICENSE: {{ .Values.global.license | b64enc }}
+  LICENSE: {{ .Values.global.license }}
+  GORILLA_LICENSE: {{ .Values.global.license }}
```
jsbroks, line 16: why did we remove this?

charts/wandb-base/templates/cronjob
```diff
@@ -0,0 +1,56 @@
+{{- range $cronJobName, $cronJob := .Values.cronJobs }}
```
jsbroks, line 1: should this faile be a yaml file?

jsbroks (approved):

charts/operator-wandb/templates/gorilla.yaml
zacharyblasczyk: When we went to test this, it was base64 encoding the base64 encoded secret. 

So removing ensured gorilla worked. 

When we initially tested with glue, it doesn't check the license.

charts/wandb-base/templates/cronjob
zacharyblasczyk: Good catch.

danielpanzella (approved):

zacharyblasczyk:
ref INFRA-234

PR #305, opened 2025-01-10
fix: we dont want the username and password for bufstream at the moment

danielpanzella:


charts/operator-wandb/values.yaml
```diff
@@ -176,6 +176,8 @@ global:
   beta:
     glue:
       enabled: false
+    bufstream:
+      enabled: true
```
zacharyblasczyk, line 180:
```suggestion
      enabled: false
```

zacharyblasczyk (approved):

PR #311, opened 2025-01-15
fix: enable flat run field updates in executor pod

jessicaxiang:
The executor pod needs to be able to write to runs v2 (flat run fields) for moving runs, etc. 

Make sure it has the proper environment variables to know which queue to write to.

### DIFF
--- a/charts/operator-wandb/charts/executor/templates/_helpers.tpl
+++ b/charts/operator-wandb/charts/executor/templates/_helpers.tpl
@@ -152,7 +152,7 @@ mysql://$(MYSQL_USER):$(MYSQL_PASSWORD)@$(MYSQL_HOST):$(MYSQL_PORT)/$(MYSQL_DATA
 {{- end -}}
 {{- end -}}
 
-{{- define "app.runUpdateShadowTopic" -}}
+{{- define "executor.runUpdateShadowTopic" -}}
 {{- if .Values.global.pubSub.enabled -}}
 pubsub:/{{ .Values.global.pubSub.project }}/{{ .Values.global.pubSub.runUpdateShadowTopic }}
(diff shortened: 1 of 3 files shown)


charts/operator-wandb/charts/executor/templates/_helpers.tpl
```diff
@@ -152,6 +152,16 @@ mysql://$(MYSQL_USER):$(MYSQL_PASSWORD)@$(MYSQL_HOST):$(MYSQL_PORT)/$(MYSQL_DATA
 {{- end -}}
 {{- end -}}
 
+{{- define "app.runUpdateShadowTopic" -}}
```
danielpanzella, line 155:
```suggestion
{{- define "executor.runUpdateShadowTopic" -}}
```

charts/operator-wandb/charts/executor/templates/deployment.yaml
```diff
@@ -126,6 +126,29 @@ spec:
...
+            {{- end }}
+            - name: GORILLA_RUN_UPDATE_SHADOW_QUEUE
+              value: >
+                {
+                  "overflow-bucket": {
+                    "store": {{ (include "wandb.bucket" . | fromYaml).url | quote}},
+                    "name": "wandb",
+                    "prefix": "wandb-overflow"
+                  },
+                  "addr": {{ include "app.runUpdateShadowTopic" . | quote }}
```
danielpanzella:
```suggestion
                  "addr": {{ include "executor.runUpdateShadowTopic" . | quote }}
```

danielpanzella (changes requested):

danielpanzella (approved):

PR #317, opened 2025-01-16
fix: Bucket postfix trailing slash

zacharyblasczyk:
This solves an issue where there is a trailing `/` in the bucket if path is not set. 

Before:
```yaml
            - name: BUCKET
              value: "gs://$(BUCKET_NAME)/$(BUCKET_PATH)"
```

After
```yaml
            - name: BUCKET
              value: "gs://$(BUCKET_NAME)"
```

The problem is that then the bucket will be

Before
```bash
BUCKET=gs://wandb-qa-local-whole-crow/
```

After
```bash
BUCKET=gs://wandb-qa-local-whole-crow
```

Which can lead to issues with the application.

### DIFF
--- a/charts/operator-wandb/templates/_bucket.tpl
+++ b/charts/operator-wandb/templates/_bucket.tpl
@@ -30,18 +30,42 @@ secretKey: {{ $secretKey }}
 accessKeyName: {{ .Values.global.bucket.secret.accessKeyName }}
 secretKeyName: {{ .Values.global.bucket.secret.secretKeyName }}
 secretName: {{ include "wandb.bucket.secret" . }}
+{{- if eq $path "" -}}
+
+{{- if eq $provider "az" -}}
+{{- $url = "az://$(BUCKET_NAME)" -}}
+{{- end -}}
+
+{{- if eq $provider "gcs" -}}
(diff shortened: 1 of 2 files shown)


dacbd (approved):

charts/operator-wandb/templates/_bucket.tpl
```diff
@@ -30,18 +30,42 @@ secretKey: {{ $secretKey }}
...
+{{- $url = "az://$(BUCKET_NAME)" -}}
+{{- end -}}
+
+{{- if eq $provider "gcs" -}}
+{{- $url = "gs://$(BUCKET_NAME)" -}}
+{{- end -}}
+
+{{- if eq $provider "s3" -}}
+{{- if or (and $accessKey $secretKey) .Values.global.bucket.secret.secretName -}}
+{{- $url = "s3://$(BUCKET_ACCESS_KEY):$(BUCKET_SECRET_KEY)@$(BUCKET_NAME)" -}}
```
vanpelt, line 45: Won't this be a problem in the self-hosted s3 case where I believe we end up setting BUCKET_NAME to the url of the s3 server?

vanpelt: I guess this is ok as path wouldn't be none in that case...

PR #322, opened 2025-01-21
feat: Backend config, mysql connection string, hpa

zacharyblasczyk:
We are planning to remove the conditional for the backend on  `{{- if .Values.global.createGCPLoadBalancerBackend }}` just need time to validate. 

We can also make the spec conditional for the backend config in the future.

ref INFRA-269

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -624,6 +624,7 @@ api:
   # install: false
   service:
     enabled: true
+    loadBalancerHealthCheckPath: "/ready"
     type: ClusterIP
     ports:
       - port: 8081
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -747,10 +747,17 @@ api:
           path: /healthz
           port: 8081
         initialDelaySeconds: 30
-        periodSeconds: 1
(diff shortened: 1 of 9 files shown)


danielpanzella (approved):

charts/operator-wandb/values.yaml
```diff
@@ -666,13 +672,29 @@ api:
...
     GORILLA_FILE_STREAM_STORE_ADDRESS:
-      value: '{{ include "wandb.mysql" . | trim }}'
+      value: '{{ include "wandb.fileStreamStoreProducer" . }}'
     GORILLA_HISTORY_STORE:
       value: 'http://{{ .Release.Name }}-parquet:8087/_goRPC_,{{ include "wandb.mysql" . | trim }}'
     GORILLA_FILE_STORE:
       value: '{{ (include "wandb.bucket" . | fromYaml).url }}'
     GORILLA_STORAGE_BUCKET:
       value: '{{ (include "wandb.bucket" . | fromYaml).url }}'
+    A_KAFKA_CLIENT_PASSWORD:
```
zacharyblasczyk, line 682: This solves an ordering issue.

danielpanzella (approved):

PR #327, opened 2025-01-24
fix: Shadow queue hard coded variables

zacharyblasczyk:


dacbd (approved):
I'm assuming the intention is they would be added as an `extraEnv` when required?

zacharyblasczyk:
> I'm assuming the intention is they would be added as an `extraEnv` when required?

Yes, the intent for this configuration at the moment should be that it is enabled on a case by case basis, not defaulted here.

PR #332, opened 2025-01-30
fix: Fix indent for annotations block under `spec`

venky-wandb:
The ident for the annotations block under `spec` is 4 instead it should be 8. This is causing the annotations to be reconciled and reverted when user sets it directly in deployment.yaml.

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -7,95 +7,95 @@ appVersion: 1.0.0
 icon: https://wandb.ai/logo.svg
 
 maintainers:
-- name: wandb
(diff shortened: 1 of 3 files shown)


flamarion (changes requested):
It is the other way around. Parquet 8 and not App 4.

venky-wandb:
Yup, jonathan is fixing it shortly.

flamarion (changes requested):
You may want to [bump the chart version](https://github.com/wandb/helm-charts/blob/venky-wandb-patch-1/charts/operator-wandb/Chart.yaml#L5) if the linter is complaining

PR #342, opened 2025-02-10
fix: Add switch to enable replicated queries on Weave Trace

flamarion:
## Summary by CodeRabbit

- **New Features**
  - Introduced a new environment variable `WF_CLICKHOUSE_REPLICATED` to manage data replication settings.
  - Updated configuration to disable replication by default, ensuring the standard storage engine is utilized.
  
- **Version Updates**
  - Helm chart version updated from `0.25.4` to `0.25.5`.


charts/operator-wandb/values.yaml
```diff
@@ -90,6 +90,8 @@ global:
       passwordKey: "CLICKHOUSE_PASSWORD"
     database: "weave_trace_db"
     user: "default"
+    # This will modify the MergeTree() engine used to insert data into the ClickHouse table to use the ReplicatedMergeTree() engine.
```
raubitsj, line 93:
```suggestion
    # Required when using replicated services on self-managed ClickHouse (not ClickHouse Cloud).
    # Flag will substitute MergeTree() with ReplicatedMergeTree() and use `ON CLUSTER` on table creation.
``

raubitsj: Suggestion since this is only needed for self-managed clickhouse

raubitsj (approved):
LGTM with one suggestion

danielpanzella (approved):

PR #346, opened 2025-02-12
chore: switch to http tracing

davidwallacejackson:
## Summary by CodeRabbit

- **Chores**
  - Updated the system’s version to 0.26.1.
- **New Features**
  - Adjusted resource allocations to better support system performance.
  - Refined the tracing service configuration by switching its communication protocol from gRPC to HTTP and updating the connection port.

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -81,9 +81,9 @@ global:
       # defaults to the otel-daemonset service
       host: ""
       # grpc default is 4317, http default is 4318
-      port: 4317
+      port: 4318
       # grpc, http
-      proto: "grpc"
+      proto: "http"
 
   clickhouse:
     install: false
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.26.0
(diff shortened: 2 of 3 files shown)


charts/operator-wandb/values.yaml
```diff
@@ -81,9 +81,9 @@ global:
       # defaults to the otel-daemonset service
       host: ""
       # grpc default is 4317, http default is 4318
-      port: 4317
+      port: 4318
```
danielpanzella, line 84: Since this will break if they are not on a very recent version of the server image, I'm hesitant to make the change here and we should instead make the change here https://deploy.wandb.ai/v2/admin/channels/9205b19d-9103-4ef6-88e3-4d7a6eda61a7 when we are sure that the server version with the code to support this is deployed fleet wide

davidwallacejackson (commented):
> Since this will break if they are not on a very recent version of the server image, I'm hesitant to make the change here and we should instead make the change here https://deploy.wandb.ai/v2/admin/channels/9205b19d-9103-4ef6-88e3-4d7a6eda61a7 when we are sure that the server version with the code to support this is deployed fleet wide 

@danielpanzella that makes sense. We roll out every release fleet-wide, right? If so I can just wait for the next release.

danielpanzella:
> @danielpanzella that makes sense. We roll out every release fleet-wide, right? If so I can just wait for the next release.

Yup, should be next week I believe.

dacbd:
stale

PR #362, opened 2025-03-08
feat: Migrate frfu

zacharyblasczyk:
## Summary by CodeRabbit

- **New Features**
  - The operator has been updated to version 0.31.0 with a refreshed dependency setup.
  - The updater is now enabled by default with new autoscaling, enhanced environment configurations, and secure service settings.
  - A flexible messaging configuration and customizable termination grace period have been introduced for smoother and more resilient deployments.
  - Added a new deployment postfix feature for enhanced naming flexibility.

- **Chores**
  - Upgraded container image source and removed outdated configurations to improve overall stability.
  - The Helm chart for `wandb-base` has been updated to version 0.7.0.

- **Documentation**
  - Added a new section for local pipeline development and troubleshooting to assist developers in running and debugging tests locally.


charts/wandb-base/values.yaml
```diff
@@ -55,6 +56,8 @@ securityContext:
   readOnlyRootFilesystem: false
   privileged: false
 
+terminationGracePeriodSeconds:
```
dacbd, line 59:
```suggestion
# https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#lifecycle
# default value is 30s
terminationGracePeriodSeconds:
```

charts/operator-wandb/values.yaml
```diff
@@ -304,15 +304,114 @@ console:
...
+#     "{{ .Release.Name }}-bucket-configmap": "configMapRef"
+
 flat-run-fields-updater:
-  install: false
-  image:
-    repository: wandb/local
-    tag: latest
-  pubSub:
-    subscription: ""
+  install: true
```
dacbd, line 318: was previously `false` should it remain false?

charts/wandb-base/templates/deployment.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: {{ include "wandb-base.fullname" . }}
+  name: {{ include "wandb-base.fullname" . }}{{ include "wandb-base.deploymentPostfix" . }}
```
dacbd, line 5: Should we add a separator `-`?

dacbd: disregard

zacharyblasczyk: Yah, did the logic in the postfix, easier that way.

charts/wandb-base/values.yaml
zacharyblasczyk: Yah, the default is 30, I just didn't want to render it if we didn't already render it. I can add the comment.

PR #366, opened 2025-03-19
feat(deployer): Redis URL construction overhaul

j7m4:
- INFRA-655 initial pass at fixing redis connection strings
- INFRA-655 fix structure from team feedback

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -148,6 +148,7 @@ global:
     # external should remain false; it is for W&B internal use only.
     external: false
     parameters: {}
+    params: {}
     caCert: ""
     secret:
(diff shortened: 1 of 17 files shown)


charts/operator-wandb/values.yaml
```diff
@@ -148,6 +148,7 @@ global:
     # external should remain false; it is for W&B internal use only.
     external: false
     parameters: {}
+    params: {}
```
zacharyblasczyk, line 151: This is because we supported both historically? I believe you just merge them, but which one takes precedence? 

I would just maybe add a comment about this if it should be deprecated.

zacharyblasczyk (approved):

danielpanzella (approved):

PR #371, opened 2025-03-28
feat: add nodeAffinity support to app deployment

flamarion:
## Summary by CodeRabbit

- **Chores**
  - Updated the version number to reflect the latest release.
  
- **New Features**
  - Introduced configurable options to define node affinity rules for pod scheduling.
  - Added flexibility to choose between preferred (soft) and required (hard) affinity strategies.
  - Provided additional configuration settings for enhanced control over deployment selection.

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.28.6
+version: 0.29.0
 appVersion: 1.0.0
 icon: https://wandb.ai/logo.svg
 
--- a/charts/operator-wandb/charts/app/templates/deployment.yaml
+++ b/charts/operator-wandb/charts/app/templates/deployment.yaml
@@ -43,6 +43,10 @@ spec:
       {{- include "wandb.nodeSelector" . | nindent 6 }}
       {{- include "wandb.priorityClassName" . | nindent 6 }}
       {{- include "wandb.podSecurityContext" .Values.pod.securityContext | nindent 6 }}
+      affinity:
+        {{- if and .Values.nodeAffinity.type .Values.nodeAffinity.key .Values.nodeAffinity.values }}
+        {{- include "wandb.nodeAffinity.type" (dict "type" .Values.nodeAffinity.type "key" .Values.nodeAffinity.key "values" .Values.nodeAffinity.values) | nindent 8 }}
+        {{- end }}
       # Extend the pods shutdown grace period from the default of 30s to 60s.
       # This goes in the pod template spec.
       terminationGracePeriodSeconds: 60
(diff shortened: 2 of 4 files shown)


flamarion:
Test results 

Helm version deployed

```bash
NAME    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART                                   APP VERSION
wandb   default         1               2025-03-28 16:50:35.43323713 +0000 UTC  deployed        operator-wandb-0.29.0-PR371-29322d3     1.0.0
```


 W&B CR config with soft node affinity


```yaml
spec:
  chart:
    url: http://charts.wandb.ai
    name: operator-wandb
    version: "0.29.0-PR371-29322d3"
  values:
    app:
      nodeAffinity:
        # Allowed values: `soft` or `hard`
        type: "hard"
        key: "fips-enabled"
        values:
          - "yes"
```


Deployment template rendered with the above YAML data:
```yaml
    spec:
      affinity:
        nodeAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            nodeSelectorTerms:
            - matchExpressions:
              - key: fips-enabled
                operator: In
                values:
                - "yes"
```

dacbd:
If we dont have any affinity settings by default we should just allow the user to set their whole own affinity block. I feel like we shouldn't add fancy helper settings that template out things like this.

flamarion:
@dacbd For those of us in the field, the more we can protect the product from being affected by incorrect configurations, the better. 
Therefore, if we can abstract everything possible and only allow customers to set true or false, that would be the ideal scenario.
Unfortunately, not every customer is familiar with Kubernetes and may not understand how to do things correctly.
In any case, if these "fancy" helpers are not good, I'm happy to close the PR :) 
Just let me know, and thank you for taking the time to review.

dacbd:
> @dacbd For those of us in the field, the more we can protect the product from being affected by incorrect configurations, the better. Therefore, if we can abstract everything possible and only allow customers to set true or false, that would be the ideal scenario. Unfortunately, not every customer is familiar with Kubernetes and may not understand how to do things correctly. In any case, if these "fancy" helpers are not good, I'm happy to close the PR. I copied the idea from Bitnami Redis, so not my invention. Just let me know, and thank you for taking the time to review.

That's fair, I'll poke more of the team to consider this.

danielpanzella:
@flamarion There is nothing you can do to prevent bad affinities from being configured, so I'm not sure its worth the overhead of protecting the user from some specific subset of bad configuration, since they can still easily add something invalid anyway.  I'd imagine its a very small subset of customers that will use this feature.

flamarion:
We should protect everything 😄 but I understand this is not doable for us now.
The Bitnami Helm Charts expose very few things for real; I mean, there is not much stuff the user can do other than set what they want.
Once again, @dacbd and @danielpanzella, thank you for taking the time to review this PR; I'm closing it.
This is not critical, so there is no rush to implement it. If I find some time, I can try to send a new PR like you suggested.

PR #373, opened 2025-03-28
fix: Allow size passthrough

dacbd:
ref: INFRA-745

## Summary by CodeRabbit

- **Chores**
  - Updated the release version to 0.28.17.
  
- **New Features**
  - Introduced a new deployment configuration option to specify scale (options include small, medium, large, xlarge, and xxlarge), with an empty string set as the default.
  - Added a new environment variable for enhanced configurability of data export parallelism in the Kubernetes CronJob specification.
  - Added a new parameter to control the level of parallelism for the export history process in the CronJob configuration.
  - Implemented dynamic configuration for export parallelism based on deployment size settings.


charts/operator-wandb/charts/parquet/values.yaml
```diff
@@ -37,6 +37,7 @@ common:
 deployment: {}
 cronJob:
   exportHistoryToParquet:
+    parallelism: "10"
```
dacbd, line 40:
```suggestion
    parallelism: ""
```

zacharyblasczyk (approved):

PR #388, opened 2025-04-16
fix(backend): history store routing in api and glue

jessicaxiang:


danielpanzella:
@jessicaxiang It looks like the test setups that include api and glue as separate pods are failing, but it's a mysql error which seem unrelated, but I'm not 100% as I think these are passing elsewhere. ```panic: Error 1045 (28000): Access denied for user 'wandb'@'10.244.0.12' (using password: YES)```  I think we should investigate and confirm its not related to your change

danielpanzella:
Ah, I see the problem.  for reasons that are complicated and stupid the envvar that holds the mysql password is call `A_MYSQL_PASSWORD` not `MYSQL_PASSWORD` to ensure that it will appear in the array before the `GORILLA_` envvars that depend on it.

danielpanzella (approved):

PR #401, opened 2025-05-07
fix: Kafka env for executor

dacbd:


levinandrew (approved):
we were getting this panic in executor pod in QA AWS
```
panic: Got error with config.Addr, parse "kafka://$(KAFKA_CLIENT_USER):wandb@$(KAFKA_BROKER_HOST):$(KAFKA_BROKER_PORT)/wandb-run-updates-shadow?producer_batch_bytes=1048576&num_partitions=12&replication_factor=3": invalid port ":$(KAFKA_BROKER_PORT)" after host
```

then adding this change directly to the user spec fixed the error, by adding `KAFKA_BROKER_PORT` env var

PR #403, opened 2025-05-09
Fix: add support for hostNetwork

jms-wandb:
Adding hostNetwork to app, buffstream, console, executor, filestream, flat-run-fields-updater, parquet, stackdriver, weave, weave-trace, and yace pods.


jms-wandb:
Going to close this out as it has diverged too much and need to start with a fresh main branch.

PR #410, opened 2025-05-15
Fix: add support for hostNetwork

jms-wandb:
Adding hostNetwork to app, buffstream, console, executor, filestream, flat-run-fields-updater, parquet, stackdriver, weave, weave-trace, and yace pods.


dacbd:
@jms-wandb, if we want to continue this line of work, it should be added to the base chart

PR #417, opened 2025-05-30
feat(WIP): Refactor all charts to use wandb-base chart

danielpanzella:


dacbd (commented):
- "legacy config" for the test-configs

PR #435, opened 2025-07-08
add values/charts updates needed for metric-observer in fmb

ssisk:
These are the helm changes needed to add the metric observer service

## Summary by CodeRabbit

* **New Features**
  * Added a configurable "metric-observer" component with install toggle, autoscaling, resources, image and env customization.
  * Added Pub/Sub subscription support for metric-observer and a new pubsub.subscription field for filestream.
  * Added global env var to enable run automations in test configs.
  * Chart dependency added for a metric-observer module and chart version bumped.

* **Improvements**
  * Reloader default changed to disable reload on resource deletion.
  * Improved environment and resource request configurability for tests.


charts/operator-wandb/values.yaml
```diff
@@ -1288,6 +1288,175 @@ frontend:
...
+    
+    # I'm pretty sure it's safe to skip these - we use these during
+    # setup of the slack integration, but by the time MO runs the slack integration,
+    # we already have all the info we need embedded in the integration
+    # GORILLA_SLACK_CLIENT_ID:
+    #    value: "{{ required "slack client id required" .Values.secretRefs.slack.alertsApp.clientId }}"
+    # GORILLA_SLACK_SECRET:
+    #    value: "{{ required "slack secret required" .Values.secretRefs.slack.alertsApp.secret }}"
+
+    # TODO - unclear if/how AZURE_STORAGE_KEY one is used
```
ssisk, line 1402: re-reading this now, I suspect this is needed for buckets, but looking at the code earlier, it wasn't obvious that it's tied to normal bucket usage

charts/operator-wandb/values.yaml
```diff
@@ -1288,6 +1288,175 @@ frontend:
...
+    #     secretKeyRef:
+    #       name: '{{ (include "wandb.bucket" . | fromYaml).secretName }}'
+    #       key: '{{ (include "wandb.bucket" . | fromYaml).accessKeyName }}'
+    #       optional: true
+    # don't believe this is used
+    # GORILLA_STATIC_HOST:
+    #   value: "{{ .Values.global.host }}"
+    
+  # I don't know how to valid envFrom/what's up with it
+  envFrom:
```
ssisk, line 1414: @danielpanzella @dacbd I realized I don't know what this is :)

charts/operator-wandb/values.yaml
```diff
@@ -1288,6 +1288,175 @@ frontend:
           cpu: 100m
           memory: 512Mi
 
+metric-observer:
+  install: false
+  imagePullSecrets: *imagePullSecrets
+  autoscaling:
```
ssisk, line 1294: I'm unclear what controls horizontal autoscaling in server environments.

charts/operator-wandb/values.yaml
```diff
@@ -1288,6 +1288,175 @@ frontend:
...
+    GORILLA_ONPREM:
+      value: "true"
+    GORILLA_FILE_STORE:
+      value: '{{ (include "wandb.bucket" . | fromYaml).url }}'
+    GORILLA_STORAGE_BUCKET:
+      value: '{{ (include "wandb.bucket" . | fromYaml).url }}'
+    GORILLA_TRACER:
+      value: '{{ include "wandb.otelTracesEndpoint" . | trim }}'
+
+    ### I'm unsure the following are necessary
```
ssisk, line 1355: these seem to be involved in otel/tracing things? I'd definitely like to have that enabled, but wasn't entirely sure what was going on with these

charts/operator-wandb/values.yaml
```diff
@@ -1288,6 +1288,175 @@ frontend:
...
+          key: '{{ (include "wandb.bucket" . | fromYaml).accessKeyName }}'
+          optional: true
+    BUCKET_SECRET_KEY:
+      valueFrom:
+        secretKeyRef:
+          name: '{{ (include "wandb.bucket" . | fromYaml).secretName }}'
+          key: '{{ (include "wandb.bucket" . | fromYaml).secretKeyName }}'
+          optional: true
+    
+    # I'm pretty sure it's safe to skip these - we use these during
```
ssisk, line 1394: I'm going to double check when we manually test that these aren't used, but I'm pretty sure they aren't

test-configs/operator-wandb/fmb.yaml
```diff
@@ -66,6 +72,34 @@ flat-run-fields-updater:
...
       memory: "128Mi"
+  image:
+    tag: 0.70.0-metric-observer-available.0
+
+metric-observer:
+  install: true
+  pubSub:
+    subscription: "metric-observer"
+  env: 
+    # TODO: I assume SHADOW_QUEUE needs to be updated on a per-env basis?
```
ssisk, line 83: this was a question for @danielpanzella @dacbd

test-configs/operator-wandb/fmb.yaml
```diff
@@ -49,6 +50,9 @@ app:
     requests:
       cpu: "100m"
       memory: "128Mi"
+  image: # TODO: DO NOT SUBMIT - remove this and all image overrides
```
ssisk, line 53: do we want to override this here? or is it easier to this in the channel?

charts/operator-wandb/values.yaml
danielpanzella: envFrom is a k8s concept that allows you to attach a ConfigMap or Secret and have all of the values within added as envvars

test-configs/operator-wandb/fmb.yaml
danielpanzella: GORILLA_RUN_UPDATE_SHADOW_QUEUE should itself be defined the same for al environments, so this should be in the values.yaml.  within it are two helm template references `{{ (include "wandb.bucket" . | fromYaml).url | quote}}` and `{{ include "wandb.runUpdateShadowQueue" . | quote }}` which will render the values appriopriate for the environment

charts/operator-wandb/values.yaml
danielpanzella: this, and possibly some other values defined here is actually included in the gorilla configMap included in your envFrom later.  only you can know what values are actually used within the application though

test-configs/operator-wandb/fmb.yaml
danielpanzella: Also, these files are just used for running tests and local dev, so they will have no bearing on the images used anywhere else.

charts/operator-wandb/values.yaml
ssisk: okay, so it sounds like it'd make sense to drop these then, so I'll see if I'm able to do that. Is there a way to see where those are defined?

test-configs/operator-wandb/fmb.yaml
ssisk: that makes sense, thanks! will make that change

charts/operator-wandb/values.yaml
ssisk: k, since we need buckets I'll plan to add that back in

test-configs/operator-wandb/fmb.yaml
ssisk: sounds good, I'll plan to remove this once I'm done with local testing before commit

test-configs/operator-wandb/fmb.yaml
ssisk: ah- I'll answer my own question. The template uses `pubsub.subscription`, and it gets that value from the metric-observer, so that's how we get the unique value. thanks!

charts/operator-wandb/values.yaml
```diff
@@ -1288,6 +1288,172 @@ frontend:
...
+      value: '{{ include "wandb.otelTracesEndpoint" . | trim }}'
+    POD_NAME:
+      valueFrom:
+        fieldRef:
+          fieldPath: metadata.name
+    G_HOST_IP:
+      valueFrom:
+        fieldRef:
+          fieldPath: status.hostIP
+    GORILLA_CUSTOMER_SECRET_STORE_K8S_CONFIG_NAMESPACE:
```
ssisk, line 1362: I'm not sure about whether we need the following 3 values, however it's extremely hard to test the gorilla megabinary to be sure that we won't need them in some corner case.

charts/operator-wandb/values.yaml
ssisk: okay, I've checked the values I could see getting set in the gorilla.yaml template for the configMap, and removed the overlaps.

charts/operator-wandb/templates/gorilla.yaml
```diff
@@ -64,7 +64,6 @@ data:
   GORILLA_ACTIVITY_STORE_SERVE: "true"
   GORILLA_GLUE_TASK_CONFIG_PATH: "/gorilla_glue_tasks_local.yaml"
   GORILLA_ONPREM: "true"
-  GORILLA_STATSD_PORT: "8125"
```
ssisk, line 67: note that this is a repeat value - see line 36

charts/operator-wandb/values.yaml
```diff
@@ -855,6 +855,8 @@ filemeta:
 filestream:
   install: false
   deploymentPostfix: "bc"
+  pubsub:
+    subscriptions: ""
```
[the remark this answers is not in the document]
ssisk: cc @danielpanzella on this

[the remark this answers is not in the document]
danielpanzella: its not for your change, I just noticed we didn't have it there, and it should be there.

test-configs/operator-wandb/__snapshots__/default.snap
```diff
@@ -1104,7 +1104,7 @@ kind: ConfigMap
 metadata:
   name: chartsnap-app-configmap
   labels:
-    helm.sh/chart: operator-wandb-0.33.0
```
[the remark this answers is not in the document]
danielpanzella: these can be regenerated with `./snapshots.sh update` but it worth glancing over the diff to ensure there are no unexpected side effects of the changes

charts/operator-wandb/values.yaml
```diff
@@ -1955,7 +2034,7 @@ reloader:
     #     tag: v1.3.0
     watchGlobally: false
     reloadOnCreate: true
-    reloadOnDelete: true
+    reloadOnDelete: false
```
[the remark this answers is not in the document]
ssisk: it's fine to leave in if it's a good thing, just wanted to make sure it wasn't a mistake

danielpanzella (approved):

PR #437, opened 2025-07-11
Online evals on dedicated deployments

neutralino1:


charts/operator-wandb/charts/weave-trace/templates/deployment-worker.yaml
```diff
@@ -0,0 +1,106 @@
...
+            timeoutSeconds: 2
+            failureThreshold: 5
+          startupProbe:
+            exec:
+              command:
+                - /bin/sh
+                - -c
+                - ps aux | grep 'scoring_worker' | grep -v 'grep'
+            timeoutSeconds: 2
+            failureThreshold: 5
```
dacbd, line 103: these probes do nothing, if the process in the command exits then the pod will get restarted

dacbd: yeah, if there is something to check where the process is still running then we can do that. but just checks to see if the process is running we don't need

charts/operator-wandb/charts/weave-trace/values.yaml
```diff
@@ -6,6 +6,13 @@ image:
   tag: latest
   pullPolicy: Always
 
+bufstream:
+  host: bufstream.bufstream.svc.cluster.local
+  port: 9092
+
+weaveWorker:
+  enabled: true
```
danielpanzella, line 14: This needs to be defaulted to false, and the test configs should be updated to test with and without this enabled

danielpanzella (changes requested):

neutralino1:
Safe to close.

PR #444, opened 2025-07-25
chore(launch): Add projected sa volume capabilities

domphan-wandb:
https://wandb.atlassian.net/browse/WB-26209

Heads up, this PR merges into the `dom/public-queue` branch which has the RBAC necessary for the launch agent to create resources for the vllm service.

This PR adds an `oidc` configuration to the chart so we can set the launch agent up as an approved launch agent (federated identity). When enabled, the `WANDB_IDENTITY_TOKEN_FILE` env var will be set to the projected service account token volume.


domphan-wandb:
> Gotta bump the chart version!

Thanks!!

bcsherma (approved):
lgtm, but I have not tested. This doesn't technically require a stamp it seems

PR #449, opened 2025-08-04
feat: Add weave-trace-worker support to refactored charts

danielpanzella:


.github/workflows/test-operator-wandb.yaml
```diff
@@ -34,29 +34,27 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
-        k8s-version: [
-            "v1.32.2",
-            "v1.31.6",
-            "v1.30.10",
-          ]
-        configuration: [
+        k8s-version: ["v1.32.2", "v1.31.6", "v1.30.10", "v1.29.14"]
```
neutralino1, line 45: This addition of `1.29.14` is not part of my PR. Is this intentional?

charts/operator-wandb/values.yaml
```diff
@@ -792,6 +797,113 @@ weave-trace:
...
+  env:
+    WF_CLICKHOUSE_PASS:
+      valueFrom:
+        secretKeyRef:
+          name: '{{ include "wandb.clickhouse.passwordSecret" . }}'
+          key: '{{ .Values.global.clickhouse.passwordSecret.passwordKey }}'
+    WF_KAFKA_BROKER_HOST:
+      value: "$(KAFKA_BROKER_HOST)"
+    WF_KAFKA_BROKER_PORT:
+      value: "$(KAFKA_BROKER_PORT)"
```
neutralino1, line 822: No longer needed. Env var names have been aligned in the code.

charts/operator-wandb/values.yaml
```diff
@@ -792,6 +797,113 @@ weave-trace:
           maxReplicas: 6
           minReplicas: 2
 
+weave-trace-worker:
```
neutralino1, line 800: So this is no longer a k8s deployment? How does this work?

neutralino1 (commented):
A few questions.

.github/workflows/test-operator-wandb.yaml
danielpanzella: I think this is an artifact of the merge that I missed.

charts/operator-wandb/templates/weave-trace.yaml
```diff
@@ -14,4 +14,32 @@ data:
   WANDB_BASE_URL: "http://{{ .Release.Name }}-app:8080/"
   {{- end }}
   WF_TRACE_SERVER_URL: "{{ .Values.global.host }}/traces"
-  WF_ENFORCE_PASSWORD_LENGTH: "false"
\ No newline at end of file
+  WF_ENFORCE_PASSWORD_LENGTH: "false"
+  WEAVE_ENABLE_ONLINE_EVAL: {{ (get  .Values "weave-trace-worker").install | quote }}
+  WEAVE_TRACE_SERVER_BASE_URL: "{{ .Values.global.host }}/traces"
+  WANDB_INTERNAL_SERVICE_TOKEN_SECRET_NAME: "weave-worker-auth"
+  KAFKA_BROKER_HOST: "bufstream.bufstream.svc.cluster.local"
```
danielpanzella, line 21: this should probably not be hard-coded in the chart.  The chart should use the values in global, and we should set this in the deployer channel.

neutralino1: I could not find the value exposed in `global`. Please advise.

danielpanzella: Its not set by default because its a config specific to dedicated.  it should probably be set by deployer channel.  If what you have here works, we can circle back to refactor that later.

PR #450, opened 2025-08-04
feat: add headless option for statefulset

annirudh:


charts/wandb-base/templates/service.yaml
```diff
@@ -30,6 +30,9 @@ metadata:
     cloud.google.com/backend-config: '{"default": "{{ include "wandb-base.fullname" . }}-backend-config"}'
     {{- end }}
 spec:
+  {{- if .Values.service.headless }}
```
danielpanzella, line 33: should we allow a headless and non-headless service to exist simultaneously

danielpanzella: or always create a normal service and optionally create the second headless service name <service_name>-headless

annirudh: The latter is better. I'll make the change.

danielpanzella: i guess we can always change this when the need comes up

danielpanzella: oh, didnt see youd already replied

annirudh: @danielpanzella updated. Does this look ok? Not sure if the right thing is to exactly copy the service attributes or pull from `service.headless`. WDYT?

PR #571, opened 2026-03-13
feat(MCP-32): Add MCP Server subchart for dedicated deployments

ash0ts:
## Summary

Adds the W&B MCP Server as an optional subchart of `operator-wandb` for dedicated/on-prem deployments. Follows the same `wandb-base` pattern used by `weave-trace`, `console`, and other services.

## Changes

- `Chart.yaml` -- new `wandb-base` dependency with alias `mcp-server`, gated by `mcp-server.install`
- `values.yaml` -- full configuration block (image, probes, resources, env vars, service)
- `templates/_mcp.tpl` -- resolves `WF_TRACE_SERVER_URL` (in-cluster weave-trace) and `WANDB_BASE_URL` (from global.host)
- `templates/_ingress.tpl` -- adds `/mcp` path routing, gated by `mcp-server.install`

## How to enable

```yaml
mcp-server:
  install: true
```

That's all a customer needs. The MCP server will be accessible at `https://wandb.customer.com/mcp`.

## Auth model

- **Now:** Clients send `Bearer <wandb-api-key>` (same as hosted)
- **Future:** OAuth 2.1 via `MCP_AUTH_MODE` env var + instance OIDC config (no chart changes needed)

## Image

`wandb/mcp-server:0.2.0` on Docker Hub, built from the private `wandb-mcp-server-test` repo. A new `publish-image.yml` workflow in that repo handles building and tagging.

## Upgrade path

Same as every other operator service: bump `mcp-server.image.tag` in values.yaml when a new version is released. Operator controller auto-reconciles.

## Jira

MCP-32

## Summary by CodeRabbit

* **New Features**
  * Optional MCP server component with configurable install, image, probes, resources and service settings.
  * Ingress routing for /mcp to the MCP server alongside existing routes.
  * In-cluster environment variables exposed to configure trace and base URLs for MCP integration.

* **Tests**
  * Added a conditional health-check test to validate MCP server readiness when enabled.

* **Chores**
  * Chart version bumped and added supporting dependency.
  * Added sample deployment configuration for MCP and related services.

---

<a href="https://app.devin.ai/review/wandb/helm-charts/pull/571" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>


ash0ts:
## QA Azure Validation (April 10, 2026)

**17/17 MCP tool tests passed** against `https://qa-azure.wandb.io/mcp` (external, through Azure Application Gateway).

### Deployment Details

- Chart version: `0.42.0-PR571-beb49fa8`
- Image: `us-docker.pkg.dev/wandb-production/public/wandb/mcp-server:0.3.0`
- Built from `wandb-mcp-server` main SHA `abba0d7` (14 tools)
- QA instance: `qa-azure.wandb.io` (Azure AKS, AGIC ingress)

### Tests Passed

| Test | Tool | Result |
|---|---|---|
| Health check | `GET /mcp/health` | 200 `{status: ok, tools_registered: 14}` |
| Auth enforcement | `POST /mcp` (no auth) | 401 |
| MCP initialize | `initialize` | protocolVersion 2025-03-26 |
| List tools | `tools/list` | 14 tools |
| Entity/projects | `query_wandb_entity_projects` | Pass |
| Count traces | `count_weave_traces_tool` | Pass |
| Infer schema | `infer_trace_schema_tool` | Pass |
| Query traces | `query_weave_traces_tool` | Pass |
| Run history | `get_run_history_tool` | Pass |
| GQL query | `query_wandb_tool` | Pass |
| Create report | `create_wandb_report_tool` | Pass (URL returned) |
| Log analysis | `log_analysis_to_wandb` | Pass |
| List registries | `list_registries_tool` | Pass |
| List artifacts | `list_artifact_versions_tool` | Pass |
| Artifact details | `get_artifact_details_tool` | Pass |
| Compare artifacts | `compare_artifact_versions_tool` | Pass |
| Docs search | `search_wandb_docs_tool` | Pass |

### Ingress Health Probe Fix

Readiness probe updated from `/health` to `/mcp/health` (commit `0caf79e8`). This matches the weave-trace pattern (`/traces/health`) and is required for Azure AGIC compatibility. Companion app.py fix in wandb-mcp-server-test PR #49.

### Reproducible Test Scripts

Test data seeded and validated using `WandBAgentFactory/developer/scripts/seed_mcp_test_data.py` and `test_mcp_deployment.py`. Same workflow works for qa-aws, qa-google, and any customer instance.

zacharyblasczyk (approved):

charts/operator-wandb/templates/_mcp.tpl
```diff
@@ -0,0 +1,64 @@
...
+
+Requires weave-trace to be installed. If weave-trace is disabled,
+override WF_TRACE_SERVER_URL in the mcp-server.env values block.
+*/}}
+{{- define "wandb.mcpEnvs" -}}
+- name: WF_TRACE_SERVER_URL
+  value: "http://{{ .Release.Name }}-weave-trace:8722"
+- name: WANDB_BASE_URL
+  value: {{ .Values.global.host | quote }}
+{{- if .Values.datadog.enabled }}
```
zacharyblasczyk, line 25: The Datadog/OTEL gates reference `.Values.datadog.enabled` and `.Values.otel.enabled` (top-level), but the chart has no top-level `datadog` key and the top-level `otel` uses `install`, not `enabled`. So both blocks are effectively dead code today, and the sibling defaults at `mcp-server.datadog.enabled` / `mcp-server.otel.enabled` in values.yaml are orphaned.

Either:
1. Scope the check under the subchart (match weave-trace at values.yaml:2851): `{{- if (index .Values "mcp-server" "datadog" "enabled") }}`, or
2. Drop the observability blocks entirely for now and add them when ddtrace/otel-distro actually ships in the image.

Mentioned in the values.yaml comment block that this is a staged rollout anyway — probably cleaner to rip it out until the image is ready.

charts/operator-wandb/templates/_mcp.tpl
```diff
@@ -0,0 +1,64 @@
...
+Resolves:
+- WF_TRACE_SERVER_URL: in-cluster weave-trace service (port 8722)
+- WANDB_BASE_URL: the W&B instance URL (from global.host)
+
+Requires weave-trace to be installed. If weave-trace is disabled,
+override WF_TRACE_SERVER_URL in the mcp-server.env values block.
+*/}}
+{{- define "wandb.mcpEnvs" -}}
+- name: WF_TRACE_SERVER_URL
+  value: "http://{{ .Release.Name }}-weave-trace:8722"
```
zacharyblasczyk, line 22: The fail-fast at line 6 tells users to escape via `mcp-server.env.WF_TRACE_SERVER_URL`, but this template still unconditionally emits `WF_TRACE_SERVER_URL` on line 21-22 pointing at the in-cluster service. If a user takes the escape hatch, they'll end up with duplicate env var entries on the pod (wandb-base typically concatenates `env` + `envTpls`), and K8s behavior on duplicates is ordering-dependent and not covered by any snapshot.

Two options:
- Gate the in-cluster emit: `{{- if not (index .Values "mcp-server" "env" "WF_TRACE_SERVER_URL") }}` so the user override wins cleanly.
- Or just drop the "or override in mcp-server.env" language from the fail message and require `weave-trace.install=true` — which the PR already implies in the docs.

charts/operator-wandb/values.yaml
```diff
@@ -2842,6 +2842,86 @@ weave:
...
+    MCP_ANALYTICS_DISABLED: "false"
+    MCP_SEGMENT_FORWARD: "true"
+    MCP_AUTH_DISABLED: "false"
+    MCP_RATE_LIMIT_ENABLED: "true"
+    MCP_RATE_LIMIT_PER_KEY: "60/minute"
+    MCP_RATE_LIMIT_GLOBAL: "1000/minute"
+    WANDB_CACHE_DIR: "/tmp/.wandb_cache"
+    WANDB_CONFIG_DIR: "/tmp/.wandb_config"
+    WANDB_DATA_DIR: "/tmp/.wandb_data"
+    HOME: "/tmp"
```
zacharyblasczyk, line 2908: The `WANDB_CACHE_DIR`/`WANDB_CONFIG_DIR`/`WANDB_DATA_DIR`/`HOME: /tmp` block looks like a workaround for non-root container filesystem perms — redirecting writes into the one writable path. If that's the case, an `emptyDir` volume mounted at the default cache path (or at `/home/wandb`) is the right primitive, not shipping fs-perm env vars as chart defaults. See what Daniel did in #499 for a similar pattern — moved the helm cache into an `emptyDir` instead of leaning on container file perms.

(Noted elsewhere: `install: false` default is intentional, so ignore that from my earlier review.)

ash0ts:
### Follow-up: runtime regression caught by manual QA testing, fix at #590

Just a cross-link for anyone following the subchart history: #590 fixes a runtime regression in \`wandb.mcpEnvs\` (_mcp.tpl) that was introduced by #588 and discovered only when we attempted a real \`mcp-server.install: true\` apply on Google QA.

Root cause: \`wandb.mcpEnvs\` is invoked via \`envTpls -> tpl . $.root\` from the subchart render path, so \`.Values\` at call time is the subchart-scoped block, not the parent values. The \`index .Values \"mcp-server\" \"...\"\` lookups added by #588 dereference a nil map at that scope.

The \`install: false\` default path (what CI tests and what every existing customer has) was never affected. The bug only fires when the mcp-server Deployment actually tries to render.

The chartsnap snap for \`mcp-server.yaml\` test config was silently capturing the helm render error as a \`kind: Unknown\` fixture and passing forever. #590 regenerates the snap (156 real resources across 17 kinds, no more error fixture) and flags the chartsnap behavior for a follow-up CI guard.

Full postmortem, reproduction commands, blast-radius analysis, and release plan for 0.42.1 in #590.

PR #572, opened 2026-03-16
chore: enable OTEL profiling

annirudh:
Enable sending profiles to the OTEL endpoint if configured.

## Summary by CodeRabbit

* **New Features**
  * Added new observability profiling endpoint configuration options in Helm values with customizable host, port, and protocol settings
  * New optional environment variable for profiler endpoint configuration when profiles are enabled

### DIFF
--- a/charts/operator-wandb/templates/_env.tpl
+++ b/charts/operator-wandb/templates/_env.tpl
@@ -578,6 +578,10 @@ Global values will override any chart-specific values.
 - name: GORILLA_TRACER
   value: '{{ include "wandb.otelTracesEndpoint" . | trim }}'
 {{- end }}
+{{- if .Values.global.otel.profiles.enabled }}
+- name: GORILLA_PROFILER
+  value: '{{ include "wandb.otelProfilesEndpoint" . | trim }}'
+{{- end }}
 {{- end -}}
 
 {{- define "wandb.sslCertEnvs" -}}
--- a/charts/operator-wandb/templates/_helpers.tpl
+++ b/charts/operator-wandb/templates/_helpers.tpl
@@ -81,6 +81,17 @@ otlp+{{ .Values.global.otel.traces.proto }}://{{ .Release.Name }}-otel-daemonset
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return the endpoint to send otel profiles to
+*/}}
+{{- define "wandb.otelProfilesEndpoint" -}}
+{{- if .Values.global.otel.profiles.host -}}
+otlp+{{ .Values.global.otel.profiles.proto }}://{{ .Values.global.otel.profiles.host }}:{{ .Values.global.otel.profiles.port }}
+{{- else -}}
+otlp+{{ .Values.global.otel.profiles.proto }}://{{ .Release.Name }}-otel-daemonset:{{ .Values.global.otel.profiles.port }}
+{{- end -}}
+{{- end -}}
+
 {{- define "wandb.internalJWTMap" -}}
 '{
 {{- $items := list -}}
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -160,6 +160,14 @@ global:
       port: 4317
       # grpc, http
       proto: "grpc"
+    profiles:
+      enabled: false
+      # defaults to the otel-daemonset service
+      host: ""
+      # grpc default is 4317, http default is 4318
+      port: 4317
+      # grpc, http
+      proto: "grpc"
 
   # Weave Trace ClickHouse connection config.
   #


charts/operator-wandb/templates/_env.tpl
```diff
@@ -578,6 +578,10 @@ Global values will override any chart-specific values.
 - name: GORILLA_TRACER
   value: '{{ include "wandb.otelTracesEndpoint" . | trim }}'
 {{- end }}
+{{- if .Values.global.otel.profiles.enabled }}
+- name: GORILLA_PROFILER
```
zacharyblasczyk, line 582: This env var is silently ignored by gorilla. `Profiler` is a `ProfilerEnv` **struct** (`conf:"profiler"`), so `segmentio/conf` maps the endpoint to the leaf field `GORILLA_PROFILER_ADDRESS` — not `GORILLA_PROFILER`. (`GORILLA_TRACER` works only because `Tracer` is a plain string.) Core's own deploy templates confirm the name.

```suggestion
- name: GORILLA_PROFILER_ADDRESS
```

zacharyblasczyk (changes requested):
Reviewed against the gorilla/megabinary consumer in `wandb/core`. One high-confidence inline fix (below), plus two things that block turning this on. I also built and validated a snapshot test — details at the end.

## 🔴 Blocking — the profiler doesn't support the `otlp+*` scheme yet (megabinary incompatibility)

`wandb.otelProfilesEndpoint` emits an `otlp+grpc://` / `otlp+http://` address, but gorilla's `SetupProfiler` only handles `noop://`, cloud profiler, and `datadog://`. Every other scheme falls through to `default:` → `fmt.Errorf("unknown profiler scheme: %s")`:

- [`setup_observability.go` L541-L584](https://github.com/wandb/core/blob/9f5ec177ecaf6c5a9b18f3b8db1e82d5b595664c/services/gorilla/cmd/setup_observability.go#L541-L584) — no `OTLPGRPCConnectionType`/`OTLPHTTPConnectionType` case (traces and metrics both have them; the profiler does not).

So `profiles.enabled=true` will make gorilla **fail at startup** on every currently-released megabinary. **This chart change is incompatible with all megabinary versions that predate a core PR adding an OTLP profiler exporter** (which isn't on `core` `main`). Please land that core change first, and gate this on a minimum megabinary version. Keeping the default `enabled: false` (👍 you did) means existing installs are safe — the risk is only for anyone who flips it on before core catches up.

> **Fleet version:** I couldn't query Datadog from my environment (no `dd` CLI / creds available here), so I can't confirm what the fleet is running. But since OTLP profiler support doesn't exist in `core` `main` at all, no deployed megabinary is compatible today regardless of fleet version — worth a quick DD check to confirm which build is live before flipping the flag anywhere.

## 🔴 Blocking — env var name (see inline suggestion)

`GORILLA_PROFILER` maps to no config field. The field is `Profiler ProfilerEnv` (a struct), so `segmentio/conf` expects the leaf as `GORILLA_PROFILER_ADDRESS`. Contrast `GORILLA_TRACER`, which works only because `Tracer` is a plain `string`:

- [`gorilla_glue.go` L140-L141](https://github.com/wandb/core/blob/9f5ec177ecaf6c5a9b18f3b8db1e82d5b595664c/services/gorilla/cmd/gorilla_glue.go#L140-L141) — `Tracer string` vs `Profiler ProfilerEnv`
- [`setup_observability.go` L80-L86](https://github.com/wandb/core/blob/9f5ec177ecaf6c5a9b18f3b8db1e82d5b595664c/services/gorilla/cmd/setup_observability.go#L80-L86) — `ProfilerEnv{ Address, Interval, HeapThreshold }`
- [`glue.yaml` L125-L128](https://github.com/wandb/core/blob/9f5ec177ecaf6c5a9b18f3b8db1e82d5b595664c/services/gorilla/deploy/gorilla/templates/glue.yaml#L125-L128) — core's own deploy templates already use `GORILLA_PROFILER_ADDRESS`

As written, the env var is silently dropped and profiling never turns on even with the flag set.

## ✅ Suggested test case (built & validated locally)

Add `test-configs/operator-wandb/otel-profiles-enabled.yaml` (copy `default.yaml`, add the `otel.profiles` block under `global:`) and regenerate with `./snapshots.sh update operator-wandb`:

```yaml
  otel:
    profiles:
      enabled: true
      host: "otel-collector.observability.svc.cluster.local"
      port: 4317
      proto: "grpc"
```

With the env-var fix applied, the snapshot renders (×4 gorilla deployments):

```yaml
- name: GORILLA_PROFILER_ADDRESS
  value: otlp+grpc://otel-collector.observability.svc.cluster.local:4317
```

⚠️ Don't add this config to the `ct install` matrix in `.github/workflows/test-operator-wandb.yaml` until core supports the OTLP profiler scheme — a live `ct install` would crash gorilla for the reason above. The chartsnap (render-only) test is safe.

zacharyblasczyk:
**Fleet version check (via Datadog)** — following up on the megabinary-compatibility point.

Queried `kubernetes.cpu.usage.total{short_image:megabinary} by {image_tag}` over the last few hours. The fleet is spread across megabinary **0.72.0 → 0.83.x** (48 distinct tags):

- Highest **stable** release running: **0.82.2** (stable tags in fleet: `0.72.0`, `0.80.0`, `0.82.0`, `0.82.2`)
- Largest bucket by minor series: **0.83.x** (mostly RC / feature builds), then 0.82.x

I then re-checked the newest core `master` (`58f9a2595c9`) — not just the commit I first reviewed — and `SetupProfiler` **still** has no `otlp+grpc`/`otlp+http` case:

- [`setup_observability.go` L528-L584 @ master](https://github.com/wandb/core/blob/58f9a2595c9/services/gorilla/cmd/setup_observability.go#L528-L584) → non-noop/cloud/datadog schemes hit `return nil, fmt.Errorf("unknown profiler scheme: %s", …)`.

**Conclusion:** no megabinary version anywhere in the fleet today (0.72 through the 0.83 RCs) can consume an `otlp+*` profiler address. Enabling `global.otel.profiles.enabled=true` would crash gorilla at startup on every currently-deployed build. This chart change needs the core OTLP-profiler-exporter work to land and ship in a megabinary release first; then the chart should document/enforce that minimum version. Default `enabled: false` keeps existing installs safe in the meantime.

PR #579, opened 2026-03-26
refactor(weave): templatize weave workers

gtarpenning:
## Summary
- Replace per-worker `wandb-base` subchart aliases (`weave-trace-worker`, `weave-evaluate-model-worker`) with a single `templates/weave-workers.yaml` that iterates over `global.weaveWorkers`.
- Adding a new worker (e.g. the alert worker) now needs only a values block + a list entry, no `Chart.yaml` dependency churn.
- Addresses [#574 review feedback](https://github.com/wandb/helm-charts/pull/574) on the weave-worker subchart explosion.

## Testing
chartsnap: rendered workers match the previous subchart output; only `weave-trace-with-worker.snap` changes.

### DIFF
--- a/charts/operator-wandb/Chart.lock
+++ b/charts/operator-wandb/Chart.lock
@@ -29,12 +29,6 @@ dependencies:
 - name: wandb-base
   repository: file://../wandb-base
   version: 0.12.0
-- name: wandb-base
-  repository: file://../wandb-base
-  version: 0.12.0
-- name: wandb-base
-  repository: file://../wandb-base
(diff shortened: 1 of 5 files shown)


gtarpenning:
Closing in favor of the subchart-alias approach in #569. Looping wandb-base per worker isn't cleanly possible (the parent can't read the subchart's value defaults, so the template would be a feature-partial fork that drops KEDA/initContainers/affinity). The per-alias pattern is what the other 24 operator-wandb components already use.

PR #585, opened 2026-04-10
feat(operator-wandb): add mailFrom to SMTP config

flamarion:
## Summary
- Adds `global.email.smtp.mailFrom` to set `GORILLA_EMAIL_FROM_ADDRESS` natively, removing the need for `global.extraEnv` workaround
- Supports both plain string values and `valueFrom` secret references, consistent with existing SMTP fields (host, port, user, password)
- Env var is only injected when `mailFrom` is non-empty, so zero impact on existing deployments

## Test plan
- [x] Snapshot test for plain string `mailFrom` (`snap-smtp-mail-from.yaml`)
- [x] Snapshot test for `valueFrom` secret ref (`snap-smtp-mail-from-secret.yaml`)
- [x] All existing snapshot tests pass (`./snapshots.sh run` — 34/34 PASS)
- [x] Default snapshot confirms `GORILLA_EMAIL_FROM_ADDRESS` is not emitted when `mailFrom` is empty

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring a global SMTP "From" address, usable as a plain value or sourced from a secret.

* **Tests**
  * Added example configurations demonstrating both direct and secret-backed sender settings.

* **Chores**
  * Bumped operator Helm chart version to 0.41.6.

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.41.5
+version: 0.41.6
 appVersion: 1.0.0
 icon: https://wandb.ai/logo.svg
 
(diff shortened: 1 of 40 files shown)


mwhooker (approved):
I think this is correct. Please fix the snapshots errors

PR #586, opened 2026-04-13
chore(backend): bufstream on and runs v2 writes on by default

harukatab:
now that dedicated customers have `bufstream` and `SHADOW_WRITES` on, set these in the helm-charts so it's on by default and it can only get disabled when the env var is explicitly set to false (avoid accidental removal of env var and stop writing to the runs v2).


charts/operator-wandb/values.yaml
harukatab, line 432: considered setting `"brokerHost": "bufstream.bufstream.svc.cluster.local"` but this should be already set from the kafka.yaml? so leaving as is.

harukatab:
this affects the onprem customer which we are not ready to roll out yet so closing this PR

https://weightsandbiases.slack.com/archives/C06CKRTPKDF/p1776209264851019?thread_ts=1776193016.540309&cid=C06CKRTPKDF

PR #588, opened 2026-04-17
fix: harden MCP subchart (address Zachary's review feedback)

ash0ts:
## Summary

Addresses all outstanding review feedback from @zacharyblasczyk on #571, plus additional issues found in a deep-dive pass of the wandb-base chart internals.

### 1. Fix dead-code datadog/otel gates (`_mcp.tpl`)

The gates referenced `.Values.datadog.enabled` (top-level) but the values are nested under `mcp-server.datadog.enabled`. Changed to `index .Values "mcp-server" "datadog" "enabled"` so the DD_* env vars actually fire when users enable datadog.

Same fix for otel gate.

### 2. Gate WF_TRACE_SERVER_URL to avoid duplicates (`_mcp.tpl`)

When a user sets `mcp-server.env.WF_TRACE_SERVER_URL` (the escape hatch for external weave-trace), the template unconditionally emitted a second `WF_TRACE_SERVER_URL` pointing to in-cluster. Now gated: only emits the in-cluster URL when the user hasn't set their own.

Note: wandb-base's `_containers.tpl` already strips envTpls duplicates when a key exists in env, so this is belt-and-suspenders.

### 3. Add emptyDir volume + podSecurityContext (`values.yaml`)

Follows the **weave-trace pattern** exactly:
- `emptyDir: {}` volume `temp-dir` mounted at `/tmp/`
- `podSecurityContext.runAsUser: 1000` to match the MCP image uid (wandb-base defaults to 999)

The existing `HOME=/tmp` and `WANDB_*_DIR=/tmp/.wandb_*` env vars correctly point into the emptyDir mount.

### 4. Add SSE proxy settings to nginx `/mcp` location (`nginx.yaml`)

MCP Streamable HTTP uses `text/event-stream` responses and long-lived `GET /mcp` SSE connections. Without `proxy_buffering off`, nginx buffers SSE events. Without adequate `proxy_read_timeout`, nginx kills idle connections.

Added:
```nginx
proxy_buffering off;
proxy_cache off;
proxy_read_timeout 86400s;
proxy_set_header Connection '';
proxy_http_version 1.1;
chunked_transfer_encoding off;
```

## Testing

- Snapshots regenerated and all pass locally
- All changes are backward-compatible (install: false default unchanged)

## References

- Zachary's review comments on #571
- weave-trace values block (pattern for emptyDir + volumeMounts)
- wandb-base `_containers.tpl` env merge logic (lines 52-66)
- wandb-base `_pods.tpl` volume handling (lines 69-111)

---

<a href="https://app.devin.ai/review/wandb/helm-charts/pull/588" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>


zacharyblasczyk (approved):

ash0ts:
### Follow-up: this PR introduced a subchart-scope regression in `wandb.mcpEnvs`

Heads-up that the \"style consistency\" rewrite of `_mcp.tpl` in this PR introduced a helm render failure on every `mcp-server.install: true` apply. Fix is up at #590 with full postmortem.

TL;DR: `wandb.mcpEnvs` is invoked via `envTpls -> tpl . $.root` from the mcp-server subchart render path, so at call time `.Values` is the subchart-scoped block (the `mcp-server:` subtree), not the full parent values. The rewrites to `index .Values \"mcp-server\" \"...\"` dereference a nil map at that scope and panic with `index of nil pointer`.

A narrower form of this bug already existed in #571 but was hidden behind `datadog.enabled: false` default. This PR made it universal by adding an unconditional `WF_TRACE_SERVER_URL` guard with the same broken access pattern.

Top-level fail-fast guard at `_mcp.tpl:1-11` was correct (parent parse scope) and is untouched. `_ingress.tpl:181` is also correct (parent render scope) and untouched. Only the `define \"wandb.mcpEnvs\"` body had to change.

Snap tests silently passed on every run since this PR merged because chartsnap captured the render error as a `kind: Unknown` fixture and matched it byte-for-byte. See #590 for details and the CI hardening follow-up.

No action needed on this PR; tracking the fix + hardening at #590.

PR #589, opened 2026-04-20
chore: remove redundant WANDB_*_DIR env vars from MCP values

ash0ts:
## Summary

Fully addresses @zacharyblasczyk's original review feedback on #571 about shipping fs-perm env vars as chart defaults.

### Removed redundant env vars

The 4 env vars (`WANDB_CACHE_DIR`, `WANDB_CONFIG_DIR`, `WANDB_DATA_DIR`, `HOME=/tmp`) are already baked into the MCP Docker image via `ENV` declarations in the Dockerfile:

\`\`\`dockerfile
ENV WANDB_CACHE_DIR=/tmp/.wandb_cache
ENV WANDB_CONFIG_DIR=/tmp/.wandb_config
ENV WANDB_DATA_DIR=/tmp/.wandb_data
ENV HOME=/tmp
\`\`\`

Shipping them as chart defaults is redundant -- the image already sets them. Zachary's original concern was "shipping fs-perm env vars as chart defaults"; this removes them.

### Kept

The emptyDir volume at `/tmp/` (added in #588) stays -- it gives the pod isolated writable scratch space matching the image's `HOME=/tmp`.

### Added: Datadog API key inheritance docs

Documented the standard W&B observability pattern in the values.yaml comment block:

- **K8s managed-install**: `mcp-server.datadog.enabled=true` -> `_mcp.tpl` sets `DD_AGENT_HOST=status.hostIP` -> node-local DD agent DaemonSet (which holds the API key from `datadog-secrets` in the `datadog` namespace) handles forwarding. **No workload-level `DD_API_KEY` needed.**
- **Cloud Run / no agent**: the `analytics.datadogApiKeySecret` block is the fallback for edge cases. Leave unset on K8s.

## Future Work (not this PR)

Longer-term cleaner fix Zachary hinted at (\"or at `/home/wandb`\"):

1. Update MCP Dockerfile: `HOME=/home/wandb`, proper `mkdir -p /home/wandb`
2. Remove `WANDB_*_DIR` env vars from Dockerfile (wandb library defaults to `$HOME/.wandb`)
3. Update helm chart emptyDir mount from `/tmp/` to `/home/wandb`

Requires new image build + chart update together. Defer to v0.3.3.

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -2911,17 +2911,27 @@ mcp-server:
     MCP_RATE_LIMIT_ENABLED: "true"
     MCP_RATE_LIMIT_PER_KEY: "60/minute"
     MCP_RATE_LIMIT_GLOBAL: "1000/minute"
-    WANDB_CACHE_DIR: "/tmp/.wandb_cache"
-    WANDB_CONFIG_DIR: "/tmp/.wandb_config"
-    WANDB_DATA_DIR: "/tmp/.wandb_data"
-    HOME: "/tmp"
-  ## Observability -- all disabled by default.
+    # Note: WANDB_CACHE_DIR, WANDB_CONFIG_DIR, WANDB_DATA_DIR, and HOME=/tmp
+    # are already set by the MCP Docker image. The emptyDir volume mounted
+    # at /tmp/ above gives the pod isolated writable scratch space matching
+    # the image's defaults.
+  ## Observability -- all disabled by default, matching other W&B services.
   ## Prerequisites before enabling:
   ##   - ddtrace / opentelemetry-distro baked into the MCP server image (>= 0.3.0)
   ##   - dd-agent DaemonSet running on nodes (for datadog)
   ##   - OTEL Collector deployed in-cluster (for otel)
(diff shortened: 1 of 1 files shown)


ash0ts:
### Retro: I saw the nil-pointer error during this PR's snap regen and dismissed it

Noting this for the record: when I ran \`./snapshots.sh update operator-wandb\` for this PR (the WANDB_*_DIR env var cleanup), the run output included a Go template error:

\`\`\`
template: operator-wandb/templates/_mcp.tpl:21:12:
  executing \"wandb.mcpEnvs\"
  at <index .Values \"mcp-server\" \"env\" \"WF_TRACE_SERVER_URL\">:
  error calling index: index of nil pointer
\`\`\`

I dismissed it as \"pre-existing noise\" because chartsnap's output immediately after said \`PASS All snapshots matched\` and the snap file already contained the error fixture (see mcp-server.snap at HEAD). That was wrong - the fail was real and had been universal since #588. One chance to catch this before 0.42.0 shipped, and I blew it.

Fix + full postmortem in #590. Follow-up CI hardening (chartsnap \`kind: Unknown\` grep guard + chart-wide \`helm template\` render gate) is tracked there.

PR #597, opened 2026-05-01
feat: dual-publish charts to OCI registry on ghcr.io

zacharyblasczyk:
## Summary

Publish changed Helm charts to `ghcr.io` as OCI artifacts while keeping the existing repository at `charts.wandb.ai`. The release job signs each OCI digest with Cosign, verifies the workflow identity, and confirms the chart can be pulled without credentials.

## Implementation

- `release.yaml`: use the pinned `chart-releaser-action` v1.7.0 commit with `skip_existing`; log in to GHCR; push, sign, and verify changed packages; then pull them with an empty registry config.
- `pr-release.yaml`: stay on the pinned v1.5.0 commit because of the [v1.6.0+ `skip_packaging` regression](https://github.com/helm/chart-releaser-action/issues/171); skip release steps when no chart changed; and stop on packaging errors.
- Pin `cosign-installer` to the v4.1.1 commit.
- Document OCI installation, Cosign verification, Helm version selection, and the `wandb/controller:1.22.0+` requirement.
- Merge current `main` and bump `operator-wandb` from 0.43.6 to 0.43.7.

## Validation

- [x] Resolve the conflicts with `main`.
- [x] Run `actionlint` on both release workflows.
- [x] Run `ct lint --config ct.yaml --charts ./charts/operator-wandb` with Helm 3.20.1.
- [x] Build dependencies and run all `operator-wandb` snapshots with Helm 3.20.1.
- [x] Run the `pr-release` workflow on the updated branch.
- [ ] Complete the Kubernetes test matrix.
- [ ] After merge, publish, sign, verify, and anonymously pull `operator-wandb:0.43.7`.
- [ ] Confirm `helm repo add wandb https://charts.wandb.ai` still works.


.github/workflows/release.yaml
```diff
@@ -41,6 +43,72 @@ jobs:
...
+        id: releaser
+        uses: helm/chart-releaser-action@cae68fefc6b5f367a0275617c9f83181ba54714f # v1.7.0
+        with:
+          skip_existing: true
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Install cosign
+        if: steps.releaser.outputs.changed_charts != ''
+        uses: sigstore/cosign-installer@dc72c7d5c4d10cd6bcb8cf6e3fd625a9e5e537da # v3.7.0
```
wandb-kc: Wouldn't be a bad idea to bump the installer SHA to match `v4.1.1`

If in the future, if we specify a `cosign-release` input version that doesn't match what's specified by the installer's default `action.yml`, we get additional checks. Claude summarized the diffs as:

- Today: install step says "this cosign binary is signed by sigstore's key"
- After bump: install step says "this cosign binary is signed by sigstore's key AND by their CI identity AND was logged to Sigstore's public log at time T"
- and a year of cosign bug fixes and the new bundle format that newer downstream tooling expects

wandb-kc:
In general, would recommend pinning the SHA for github actions earlier in the workflow too but could be a different PR

j7m4 (approved):
For some of the lengthy inline scripts, I would recommend extracting them out into shell scripts, like we do in the deployments repo. I don't see it as necessary for this PR, however.

PR #598, opened 2026-05-08
chore(weave): mount internal JWT + allowlist worker SA on gorilla

amwarrier:
weave-trace-worker historically reached gorilla only indirectly: it called weave-trace's HTTP endpoints (e.g. POST /completions/create), and weave-trace was the one talking to gorilla on the worker's behalf. That worked but coupled the worker's throughput and tail latency to weave-trace's HTTP request handling — every scoring eval took a round-trip through the trace-server's FastAPI loop, even when the worker had everything it needed to drive the LLM call, fetch secrets, and write to ClickHouse on its own.

To let the worker do that work directly (and shed the round-trip), the worker now authenticates to gorilla itself for endpoints that require service identity — notably /service-dangerzone/secrets/ {entity} for custom-provider API-key fetches. The chart was missing the two pieces required to make that work:

  1. weave-trace-worker.volumes / containers.weave-trace-worker. volumeMounts: a projected serviceAccountToken volume mounted at /tmp/weave-trace/internal-jwt, audience internal-service, expirationSeconds 600 (kubelet rotates). The worker reads /tmp/weave-trace/internal-jwt/token and sends it as the Gorilla-Internal-JWT header. Same shape as weave-trace's existing mount.

  2. app.internalJWTMap: a second allowlist entry for the worker's SA (system:serviceaccount:<ns>:<release>-weave-trace-worker) pointing at the same in-cluster K8s issuer. Without this, gorilla rejects the JWT with 401.

After this change, weave-trace-worker keeps its default SA, talks to gorilla as itself, and no longer requires weave-trace as a service-to-service proxy. Audit logs show distinct callers, and the two services' RBAC can diverge cleanly as the worker's responsibilities grow.

## Summary by CodeRabbit

* **Chores**
  * Enhanced internal service authentication for the weave-trace-worker by adding projected service-account JWT support (mounted token with audience "internal-service" and 600s expiry) to enable secure service-to-service calls.
  * Bumped the Helm chart version to 0.42.5.

[Review Change Stack](https://app.coderabbit.ai/change-stack/wandb/helm-charts/pull/598)

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.42.4
+version: 0.42.5
 appVersion: 1.0.0
 icon: https://wandb.ai/logo.svg
 
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -737,6 +737,8 @@ app:
   internalJWTMap:
     - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace"
       issuer: "https://kubernetes.default.svc.cluster.local"
+    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace-worker"
+      issuer: "https://kubernetes.default.svc.cluster.local"
   env:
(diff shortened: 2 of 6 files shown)


gtarpenning (approved):
lgtm from weave team

zacharyblasczyk (approved):
CC @danielpanzella re the comment [here](https://github.com/wandb/helm-charts/pull/598/changes#diff-49582d692acfce2358472a39bb9148cf4e8f2c637f814b07665c7d3d8797933aR734)

> TODO(dpanzella) Read from here for backwards compatability, needs to be moved so API can access as well

PR #599, opened 2026-05-20
add history reader service

peytonsarmiento:
## Summary by CodeRabbit

* **New Features**
  * Added an optional ClickHouse-backed `history-reader` gRPC service for faster run history lookups.
  * When enabled, gorilla history queries prefer `history-reader` before existing parquet/mysql/bigtable fallbacks.
  * New configuration for gRPC message limits, in-memory cache (size/TTL), health checks, and autoscaling tiers.
* **Chores**
  * Updated the `operator-wandb` Helm chart version to `0.42.8` and wired in the `history-reader` chart dependency.
* **Style**
  * Updated Datadog autodiscovery tag rendering to use double-quoted YAML strings.


charts/operator-wandb/values.yaml
```diff
@@ -1966,6 +1969,141 @@ history-updater:
                   selectPolicy: Max
           # triggers must be provided by user when enabling KEDA
 
+# history-reader is a gRPC service that serves run history from ClickHouse.
+history-reader:
+  install: false
+  image:
+    repository: wandb/megabinary
+    tag: latest
+  service:
```
robertquitt, line 1978: Doesn't the service need to have a name? Or does it have a default

robertquitt: Might be worth adding a snapshot test to test this

peytonsarmiento: It has a default of `wandb-base.fullName` defined in [service.yaml](https://github.com/wandb/helm-charts/blob/main/charts/wandb-base/templates/service.yaml#L64) which is created from the release name + chart name in a [helper ](https://github.com/wandb/helm-charts/blob/main/charts/wandb-base/templates/_helpers.tpl#L13). It looks like the default for the other services is to rely on `wandb-base.fullName` so I think it makes sense to use that as well. I can add a snapshot test though! It would be good to test this

charts/operator-wandb/values.yaml
```diff
@@ -1966,6 +1969,141 @@ history-updater:
...
+          enabled: true
+          maxReplicas: 2
+          minReplicas: 2
+    medium:
+      resources:
+        limits:
+          cpu: "6"
+          memory: 16Gi
+        requests:
+          cpu: "6"
```
jsingleton-dev: That's a lot of CPU. You sure you need that much for medium? Normally you'd request much lower and have an HPA that scales up your pods as needed based on CPU usage

charts/operator-wandb/values.yaml
```diff
@@ -1966,6 +1969,141 @@ history-updater:
...
+          memory: 16Gi
+      autoscaling:
+        horizontal:
+          enabled: true
+          maxReplicas: 2
+          minReplicas: 2
```
jsingleton-dev, line 2112: You sure you want a static number of pods for all sizes? You know the workload better than me but seems like being able to scale up would be something you'd want

jsingleton-dev (approved):

charts/operator-wandb/values.yaml
peytonsarmiento: I had initially provisioned it higher because we are doing history reads and the pods that do that normally use more CPU. We just did some load testing though and I believe since it's CH instead of parquet, we use less CPU so I can lower this! I'm thinking to 2

charts/operator-wandb/values.yaml
peytonsarmiento: Ah this was a copy paste error, thanks for catching it!

PR #606, opened 2026-06-04
feat(operator-wandb): use deployment-wide Azure storage identity

zacharyblasczyk:
## Summary

This chart adds a deployment-wide Azure storage identity while preserving existing key-based and legacy BYOB behavior.

- `global.azureStorageIdentity` holds the deployment tenant, client, principal, and shared ServiceAccount contract
- the chart creates one `wandb-bucket-access` ServiceAccount by default, or uses an externally created account with `create: false`
- `wandb-base`, Weave, and Bufstream storage consumers select the shared account and keep their RoleBindings aligned
- literal client IDs are applied to the ServiceAccount annotation automatically; reference-valued client IDs require an explicit literal annotation or an external account
- `global.bucket.azureAuthMethod` lets an Azure customer bucket select `accessKey` or `workloadIdentity`
- existing bucket-scoped tenant/client IDs and explicit storage-account keys remain supported
- both Azure credential fields are omitted from the Secret when workload identity is selected

Weave file storage remains off by default. `global.weave-trace.fileStorage.useDefaultBucket` maps only the managed `global.defaultBucket` to `WF_FILE_STORAGE_URI`; it rejects bucket paths containing an object prefix. Mixed access-key BYOB plus Weave default-bucket identity is covered explicitly.

Grant the deployment identity `Reader` and `Storage Blob Data Contributor` on the storage account. Existing Azure BYOB continues to use storage-account keys unless it explicitly selects `workloadIdentity`.

The operator chart remains at feature release `0.44.0`.

## Verification

- focused Azure storage Helm suite: 25 passed
- template formatter and 16 Python tests
- template maintainability check
- `ct lint`
- full operator-wandb snapshot suite with Helm 3.20.1

## Azure QA validation

Validated on 2026-07-30 with chart prerelease `0.44.0-PR606-13c71a2b` and Core prerelease `0.84.0-azure-sas-blob-test.1`.

- W&B SDK run files and artifacts used Azure user-delegation SAS for GET and PUT
- sibling blob access returned HTTP 403
- allowlisted Weave storage wrote to the managed Azure default bucket
- non-allowlisted Weave storage remained inline
- disabling the feature restored the original account-key SAS behavior; every `wandb verify` check passed on the restored runtime

The final follow-up commit adds regression coverage only; it does not change rendered chart output.

## Rollout

Release this chart before enabling wandb/core#49026. Keep the Core migration disabled for existing installations until they are explicitly selected for rollout.

Jira: WB-38011

Companion PRs: wandb/core#49026, wandb/docs#2967

## Summary by CodeRabbit

* **New Features**
  * Added Azure Storage workload identity support, including shared service accounts, credential references, and automatic workload identity labeling.
  * Added configuration options for Azure tenant/client IDs and access-key or workload-identity authentication.
  * Added Azure BYOB and workload identity deployment configurations.
  * Improved Weave Trace storage URI handling for Azure-backed storage.

* **Bug Fixes**
  * Prevented unnecessary Azure access keys and secrets from being rendered when workload identity is enabled.

* **Tests**
  * Added comprehensive coverage for Azure authentication, service accounts, RBAC, and storage configurations.

### DIFF
--- a/charts/operator-wandb/templates/_bucket.tpl
+++ b/charts/operator-wandb/templates/_bucket.tpl
@@ -13,6 +13,31 @@
 {{ .Release.Name }}-bucket-configmap
 {{- end -}}
 
+{{/*
+Returns the effective Azure storage identity. The dedicated global override is
+owned by the deployment and takes precedence over the CR/user bucket values.
+*/}}
+{{- define "wandb.azureStorageIdentity" -}}
+{{- $bucket := include "wandb.bucket" . | fromYaml -}}
+{{- $override := default (dict) .Values.global.azureStorageIdentity -}}
+{{- $overrideTenantId := default "" $override.tenantId -}}
+{{- $overrideClientId := default "" $override.clientId -}}
+{{- $tenantId := default "" $bucket.azureTenantId -}}
+{{- $clientId := default "" $bucket.azureClientId -}}
+{{- $overrideConfigured := or (not (empty $overrideTenantId)) (not (empty $overrideClientId)) -}}
+{{- if $overrideConfigured -}}
+  {{- if ne (empty $overrideTenantId) (empty $overrideClientId) -}}
+    {{- fail "global.azureStorageIdentity.tenantId and clientId must be provided together" -}}
+  {{- end -}}
+  {{- $tenantId = $overrideTenantId -}}
+  {{- $clientId = $overrideClientId -}}
+{{- else if and (eq $bucket.provider "az") (ne (empty $tenantId) (empty $clientId)) -}}
+  {{- fail "Azure bucket azureTenantId and azureClientId must be provided together" -}}
+{{- end -}}
+{{- $enabled := and (eq $bucket.provider "az") (not (empty $tenantId)) (not (empty $clientId)) -}}
+{{- dict "enabled" $enabled "tenantId" $tenantId "clientId" $clientId | toYaml -}}
+{{- end }}
+
 
 {{- define "wandb.bucket" -}}
 {{- $url := "" -}}
--- a/charts/operator-wandb/templates/_bucket.tpl
+++ b/charts/operator-wandb/templates/_bucket.tpl
@@ -72,8 +72,8 @@ provider: {{ $provider }}
 path: {{ $path }}
 accessKey: {{ $accessKey }}
 secretKey: {{ $secretKey }}
-azureTenantId: {{ $azureTenantId }}
-azureClientId: {{ $azureClientId }}
+azureTenantId: {{ $azureTenantId | toJson }}
+azureClientId: {{ $azureClientId | toJson }}
 accessKeyName: {{ .Values.global.bucket.secret.accessKeyName }}
 secretKeyName: {{ .Values.global.bucket.secret.secretKeyName }}
 secretName: {{ include "wandb.bucket.secret" . }}
--- a/charts/operator-wandb/templates/_bucket.tpl
+++ b/charts/operator-wandb/templates/_bucket.tpl
@@ -15,7 +15,8 @@
 
 {{/*
 Returns the effective Azure storage identity. The dedicated global override is
-owned by the deployment and takes precedence over the CR/user bucket values.
+supplied by the customer's infrastructure and takes precedence over the
+W&B-managed CR bucket values.
 */}}
 {{- define "wandb.azureStorageIdentity" -}}
 {{- $bucket := include "wandb.bucket" . | fromYaml -}}
@@ -36,6 +37,11 @@ owned by the deployment and takes precedence over the CR/user bucket values.
 {{- end -}}
 {{- $enabled := and (eq $bucket.provider "az") (not (empty $tenantId)) (not (empty $clientId)) -}}
 {{- dict "enabled" $enabled "tenantId" $tenantId "clientId" $clientId | toYaml -}}
+{{- end }}
+
+{{/* Return whether the effective Azure storage identity is configured. */}}
+{{- define "wandb.azureStorageIdentityEnabled" -}}
+{{- (include "wandb.azureStorageIdentity" . | fromYaml).enabled -}}
 {{- end }}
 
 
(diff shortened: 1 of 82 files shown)


charts/operator-wandb/templates/_env.tpl
```diff
@@ -53,12 +53,29 @@ Global values will override any chart-specific values.
...
+{{- $azureIdentity := include "wandb.azureStorageIdentity" . | fromYaml -}}
+{{- if not $azureIdentity.enabled }}
 - name: AZURE_STORAGE_KEY
   valueFrom:
     secretKeyRef:
       name: {{ (include "wandb.bucket" . | fromYaml).secretName | quote }}
       key: {{ (include "wandb.bucket" . | fromYaml).accessKeyName | quote }}
       optional: true
+{{- end }}
+{{- if $azureIdentity.enabled }}
```
dacbd: we can have style discussion, is there more precedent in the chart for this being `two if's` or an `if else`

charts/operator-wandb/templates/_env.tpl
```diff
@@ -53,12 +53,29 @@ Global values will override any chart-specific values.
...
+{{- else }}
+  value: {{ $azureIdentity.tenantId | quote }}
+{{- end }}
+- name: AZURE_STORAGE_CLIENT_ID
+{{- if kindIs "map" $azureIdentity.clientId }}
+{{- toYaml $azureIdentity.clientId | nindent 2 }}
+{{- else }}
+  value: {{ $azureIdentity.clientId | quote }}
+{{- end }}
+{{- end }}
```
dacbd: can we add whitespace on the control block to make it a bit more human readable?

```suggestion
{{- if $azureIdentity.enabled }}
- name: AZURE_STORAGE_TENANT_ID
  {{- if kindIs "map" $azureIdentity.tenantId }}
    {{- toYaml $azureIdentity.tenantId | nindent 2 }}
  {{- else }}
  value: {{ $azureIdentity.tenantId | quote }}
  {{- end }}
- name: AZURE_STORAGE_CLIENT_ID
  {{- if kindIs "map" $azureIdentity.clientId }}
    {{- toYaml $azureIdentity.clientId | nindent 2 }}
  {{- else }}
  value: {{ $azureIdentity.clientId | quote }}
  {{- end }}
{{- end }}
```

charts/operator-wandb/templates/_bucket.tpl
```diff
@@ -41,6 +140,8 @@ provider: {{ $provider }}
 path: {{ $path }}
 accessKey: {{ $accessKey }}
 secretKey: {{ $secretKey }}
+azureTenantId: {{ $azureTenantId | toJson }}
+azureClientId: {{ $azureClientId | toJson }}
```
dacbd, line 162: is `toJson` correct, can you add a snapshot test that shows this, or highlight an existing snapshot that shows what this value looks like?

charts/operator-wandb/values.yaml
```diff
@@ -577,6 +613,8 @@ artifactsGc:
   DeleteFilesNumWorkers: 0
 
 api:
+  podLabels:
+    azure.workload.identity/use: '{{ include `wandb.azureStorageIdentityEnabled` . }}'
```
dacbd, line 624: for these, does the service account need an annotation too?

charts/operator-wandb/values.yaml
```diff
@@ -3444,6 +3508,8 @@ mcp-server:
 weave-trace:
   install: false
   deploymentPostfix: "bc"
+  podLabels:
+    azure.workload.identity/use: '{{ include `wandb.weaveTraceUsesAzureWorkloadIdentity` . }}'
```
dacbd, line 3824: Like the above will the weave-trace service account need an annotations point that service account at a different identity, iirc there is something in the application that try to write to the secret store?

charts/operator-wandb/values.yaml
zacharyblasczyk: Yes. The pod label opts the pod into Azure Workload Identity mutation; its Kubernetes ServiceAccount also needs the `azure.workload.identity/client-id` annotation and a matching federated credential. For W&B-managed Azure installs, Core already supplies this on `api.serviceAccount.annotations` in `services/deployer/terraform/deploy/azure/single-tenant/module/main.tf`; wandb/core#49026 keeps that wiring and extends the shared ServiceAccount/federation map. The Helm chart leaves this infrastructure-owned because `global.azureStorageIdentity.clientId` may be an EnvVarSource/secret reference, which cannot be rendered as a scalar metadata annotation. Standalone installs must set `api.serviceAccount.annotations` and federation separately.

charts/operator-wandb/templates/_env.tpl
zacharyblasczyk: Addressed by the subsequent helper refactor; the nested conditional no longer exists in this form, and the rendered behavior is covered by the focused Azure storage authentication suite.

charts/operator-wandb/templates/_env.tpl
zacharyblasczyk: Addressed by the helper refactor and formatter pass. The current template structure no longer uses this dense control-block layout.

charts/operator-wandb/templates/_bucket.tpl
zacharyblasczyk: Confirmed. The reference-valued tenant/client IDs are serialized through the helper and rendered back as valueFrom blocks. The azure-workload-identity-refs snapshot and focused Helm test now cover the exact output, including the required literal ServiceAccount annotation for a reference-valued client ID.

charts/operator-wandb/values.yaml
zacharyblasczyk: Addressed. Weave and its worker variants select the deployment-wide wandb-bucket-access ServiceAccount only when useDefaultBucket enables Azure workload identity. The shared account carries the Azure client-ID annotation, and regression coverage includes the mixed access-key BYOB/default-bucket case.

zacharyblasczyk:
@coderabbitai  full review.

mleonidas (approved):

PR #609, opened 2026-06-08
chore(weave): add weave trace agent scoring worker dedicated

jtschoonhoven:
## Summary

- Add the weave-trace agent scoring worker as a `wandb-base` subchart alias, matching how every other operator-wandb worker is deployed.
- Mirrors the prod chart [weave-agent-scoring-worker-deployment.yaml](https://github.com/wandb/core/blob/master/services/weave-trace/charts/weave-trace/templates/weave-agent-scoring-worker-deployment.yaml) and follows the same pattern as the alert worker in #569.
- Runs `python -m src.workers.agent_scoring_worker` (a Kafka consumer, so no HTTP port/service).
- Wires `WEAVE_ENABLE_AGENT_SCORING` into the shared `weave-trace-configmap`, gated on `weave-trace-agent-scoring-worker.install`. Both the trace server and the worker read this flag (prod sets it in `weave-trace-config.yaml` and `weave-agent-scoring-worker-config.yaml`), matching the existing `WEAVE_ENABLE_ONLINE_EVAL` pattern.

## Notes / assumptions

- Resource sizing mirrors the alert worker from #569 (6Gi limit, 1 CPU / 3Gi request). Prod sets agent-scoring resources per-environment, not in the chart defaults, so adjust if a heavier profile is wanted.
- `install: false` by default; enabled in the `weave-trace-with-worker` snapshot config.

## Testing

- [x] Tested image locally
- [x] Tested in QA

Deployed chart to QA and confirmed the agent receives bufstream events:

Screenshot 2026-06-11 at 11 35 23 AM

## Summary by CodeRabbit

* **New Features**
  * Added an optional agent scoring worker component that can be enabled via configuration, with autoscaling support and preset resource sizes for different deployments; includes a configurable toggle exposed to deployment settings.

* **Chores**
  * Bumped Helm chart version to 0.42.7.

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -3397,6 +3397,93 @@ weave-evaluate-model-worker:
           maxReplicas: 6
           minReplicas: 2
 
+weave-trace-agent-scoring-worker:
+  install: false
+  deploymentPostfix: "bc"
+  image:
+    repository: wandb/weave-trace
+    tag: latest
+  envFrom:
+    "{{ .Release.Name }}-global-secret": "secretRef"
+    "{{ .Release.Name }}-weave-trace-configmap": "configMapRef"
+  envTpls:
+    - '{{ include "wandb.clickhouseConfigEnvs" . }}'
+    - '{{ include "wandb.sslCertEnvs" . }}'
+  env:
+    WANDB_INTERNAL_SERVICE_TOKEN:
+      valueFrom:
+        secretKeyRef:
+          name: weave-worker-auth
+          key: key
+  containers:
+    weave-trace-agent-scoring-worker:
+      args:
+        - "python"
+        - "-m"
+        - "src.workers.agent_scoring_worker"
+  size: ""
+  sizing:
+    small:
+      resources:
+        limits:
+          memory: 6Gi
+        requests:
+          cpu: "1"
+          memory: 3Gi
+      autoscaling:
+        horizontal:
+          enabled: true
+          maxReplicas: 2
+          minReplicas: 1
+    medium:
+      resources:
+        limits:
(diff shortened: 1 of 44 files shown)


charts/operator-wandb/values.yaml
```diff
@@ -3397,6 +3397,93 @@ weave-evaluate-model-worker:
...
+        - "-m"
+        - "src.workers.agent_scoring_worker"
+  size: ""
+  sizing:
+    small:
+      resources:
+        limits:
+          memory: 6Gi
+        requests:
+          cpu: "1"
```
jsingleton-dev, line 3431: How did you determine the 1 CPU requirement? Any QA or loadtest runs where we can see the normal expected usage?

jsingleton-dev (approved):
Left a non-blocking question. Looks good let's just make sure we aren't over requesting CPU

charts/operator-wandb/values.yaml
jtschoonhoven: Thanks for the review!

I copied this from Griffin's alert worker PR https://github.com/wandb/helm-charts/pull/569/changes because the workloads are similar and because this worker should be entirely I/O bound.

It's not very high-consequence if we find out we need to adjust this later on.

jsingleton-dev: Yeah keep an eye on it in prod and adjust the requests accordingly. It's not a bunch of pods so not that big of a deal but if we're carving out 1 CPU of requests but only using 100m, it's a lot of wasted compute. Requests are the min you need to function k8s will let you burst into more if you need it, so generally good to keep requests pretty low to save on compute costs. Probably something we need to review more broadly on all of our apps though

PR #612, opened 2026-06-11
[DO NOT MERGE]: Testing CI Pipeline

jsingleton-dev:
Looks like ci pipeline may be broken on master. Testing with this no-op pr

## Summary by CodeRabbit

* **Chores**
  * Updated development configuration file with minor adjustments.

### DIFF
--- a/charts/operator-wandb/charts/clickhouse/values.yaml
+++ b/charts/operator-wandb/charts/clickhouse/values.yaml
@@ -166,8 +166,8 @@ defaultInitContainers:
 ##
 image:
   registry: docker.io
-  repository: bitnamilegacy/clickhouse
-  tag: 25.7.5-debian-12-r0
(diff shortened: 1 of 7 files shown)


charts/operator-wandb/charts/clickhouse/values.yaml
```diff
@@ -166,8 +166,8 @@ defaultInitContainers:
 ##
 image:
   registry: docker.io
-  repository: bitnamilegacy/clickhouse
-  tag: 25.7.5-debian-12-r0
```
jsingleton-dev, line 170: Testing this one out because it's what operator v2 is using: https://github.com/wandb/operator/blob/80f51a441d6ed56765b80f618e66c3c036df0d0f/internal/controller/infra/managed/clickhouse/altinity/spec.go#L24

PR #631, opened 2026-06-26
fix: make clickhouse olap variables optional

ibindlish:
<Developed with the help of Claude>

On dedicated/on-prem installs, the OLAP (ClickHouse) connection env vars (registry search / run-store accelerator / history) are built from external K8s refs (configMapKeyRef/secretKeyRef). These were emitted without optional, so if a referenced Secret/ConfigMap is missing (e.g. the ClickHouse password secret isn't provisioned on an instance), the kubelet can't resolve the env var and the pod fails to schedule (CreateContainerConfigError) — before gorilla even starts. That bricks the whole api/glue pod, not just registry search.

This complements the gorilla-side change that makes the registry-search ClickHouse connection a no-op when the address is invalid — but that graceful degradation only helps if the pod actually schedules.
Hence this helm fix.

### Change: 
- New helper wandb.olapFieldRef (_olap.tpl): renders an OLAP env field and injects optional: true into any secretKeyRef/configMapKeyRef.
- _env.tpl (wandb.olapFeatureEnvs): routes the valueFrom-map fields (host/port/database/user/password) through that helper.

### Tests
- New fixture test-configs/operator-wandb/olap-external-refs.yaml exercises the external-ref path
- __snapshots__/olap-external-refs.snap asserts optional: true lands on every olap ref.

Tested on QA-Google
Chart: 0.42.8-PR631-a4c9a96b
Deleted the clickhouse secret and saw the container get scheduled. Then gorilla was able to start up as expected.


zacharyblasczyk:
I think this change belongs at a different part of the stack. Let's discuss internally.

PR #636, opened 2026-07-07
feat: opt-in leader election for gorilla migrate initContainers

vas28r13:
## What

Adds opt-in leader election so concurrent `migrate-db` initContainers (api, glue)
serialize DB migrations via a Kubernetes Lease instead of racing.

## How

- New `leader-election-rbac.yaml` template in `wandb-base` — any component can set
  `leaderElection.enabled: true` to grant its ServiceAccount the lease permissions
  (scoped to `leaseName` where possible)
- `api` / `glue` wire `GORILLA_LEADER_ELECTION_*` env into `migrate-db`, driven by
  the same flag — one switch enables both RBAC and behavior
- Operator role gains `coordination.k8s.io/leases` so it can delegate the grant
  (RBAC escalation prevention)

Off by default; no behavior change unless enabled. Requires megabinary with
leader-election support in `migrate`.

## Summary by CodeRabbit

* **New Features**
  * Added leader election support for select services, including configurable lease settings and related permissions.
  * Introduced updated chart versions and dependency refreshes across multiple components.

* **Bug Fixes**
  * Improved deployment reliability by ensuring required lease access is available when leader election is enabled.

* **Tests**
  * Added new test configurations covering leader election and operator deployment scenarios.

### DIFF
--- a/charts/lumen/Chart.lock
+++ b/charts/lumen/Chart.lock
@@ -1,9 +1,9 @@
 dependencies:
 - name: wandb-base
   repository: file://../wandb-base
-  version: 0.12.3
+  version: 0.12.4
 - name: wandb-base
   repository: file://../wandb-base
-  version: 0.12.3
-digest: sha256:f996aeb5449178585664e12a9d53cc5ed2e2830fc6fc1561f6c4c0c8afde4d0e
-generated: "2026-07-06T12:05:12.893621-05:00"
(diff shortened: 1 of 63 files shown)


vas28r13:
Pivoting to a cleaner approach of a post-install, pre-upgrade job for CH migrations: https://github.com/wandb/helm-charts/pull/638

PR #638, opened 2026-07-08
feat: run ClickHouse migrations as a helm hook job instead of initContainers

vas28r13:
## What

Adds an opt-in `clickhouseMigrationJob` that runs `megabinary migrate` once per
helm operation as a hook Job, so concurrent api/glue pods no longer race to run
ClickHouse migrations (ClickHouse has no DB-level locking, unlike MySQL).

## How it works

- Runs as `post-install` (after in-release ClickHouse exists) and `pre-upgrade`
  (migrations finish before new pods roll)
- A wait initContainer blocks until ClickHouse accepts connections; the migrate
  container then runs exactly once — no retries on real failures (`backoffLimit: 0`),
  and a hung migration can't block the release (`activeDeadlineSeconds: 900`)
- Existing `migrate-db` initContainers are unchanged and become fast no-ops on
  upgrades; they remain the safety net for fresh installs and envs without the hook

Off by default — enable with `clickhouseMigrationJob.install: true`.
No operator chart or RBAC changes needed.

## Summary by CodeRabbit

* **New Features**
  * Added an optional ClickHouse migration Job to Helm chart deployments.
  * Introduced new Helm values to control the Job (install toggle, image overrides, and timeout).
  * Expanded test configuration to cover a deployment enabling the migration Job and related components.

* **Bug Fixes**
  * Improved migration handling so missing migrations don’t cause Job failures in supported cases.

* **Chores**
  * Bumped the Helm chart version to **0.43.3**.


charts/operator-wandb/values.yaml
```diff
@@ -2939,6 +2939,101 @@ settingsMigrationJob:
...
+# var is unset, so the job is a fast no-op unless OLAP features are enabled;
+# MySQL schemas stay with the initContainers. The image must carry the same
+# migration files as the deployed code -- keep its tag in lockstep with api.
+# TODO: default to true once the deployer stamps this job's image tag
+# alongside api's (wandb/deployments user-spec/deploy.py).
+clickhouseMigrationJob:
+  install: false
+  kind: Job
+  globalOptOut:
+    volumes: true
```
dacbd: iirc this might cause issues with custom cert bundles

charts/operator-wandb/values.yaml
```diff
@@ -2939,6 +2939,101 @@ settingsMigrationJob:
...
+    hook: "post-install,pre-upgrade"
+  envTpls:
+    - '{{ include "wandb.sslCertEnvs" . }}'
+    - '{{ include "wandb.historyEnvs" . }}'
+    - '{{ include "wandb.registrySearchEnvs" . }}'
+    - '{{ include "wandb.runStoreAcceleratorEnvs" . }}'
+  jobs:
+    ch-migrate:
+      # Only render when at least one global.olap feature is enabled.
+      enabled: '{{ include "wandb.olapAnyFeatureEnabled" . }}'
```
dacbd, line 2974: 👍 good workaround

charts/wandb-base/templates/job.yaml
```diff
@@ -26,8 +30,11 @@ spec:
   {{- with $job.ttlSecondsAfterFinished }}
   ttlSecondsAfterFinished: {{ . }}
   {{- end }}
-  {{- with $job.backoffLimit }}
-  backoffLimit: {{ . }}
+  {{- if hasKey $job "backoffLimit" }}
+  backoffLimit: {{ $job.backoffLimit }}
+  {{- end }}
+  {{- with $job.activeDeadlineSeconds }}
+  activeDeadlineSeconds: {{ . }}
```
dacbd: why change backoffLimit but not these other 2?

dacbd (approved):

charts/operator-wandb/values.yaml
vas28r13: good catch - removing

charts/wandb-base/templates/job.yaml
vas28r13: fixed it up a bit but 0 is valid for `backoffLimit` and `ttlSecondsAfterFinished ` but not for `activeDeadlineSeconds`

PR #639, opened 2026-07-08
fix(parquet): right-size medium+ CPU request to fit the 16-vCPU node pool (WB-36925)

zacharyblasczyk:
## Why

`WB-36925` (P2): fresh/upsized single-tenant installs at **medium and larger** come up with `wandb-parquet` stuck `Pending` and its rollout wedged. This is not a quota or node-count problem — it's a sizing-contract mismatch between this chart and the deployer that no amount of scaling can fix.

The two sizing ladders are set from the same t-shirt size but never reconciled:

- **This chart** requests **15 CPU** (`request == limit`) for `parquet` at `medium`/`large`/`xlarge`/`xxlarge`.
- **The deployer** provisions a **16-vCPU node** for *every* size from `medium` through `xxlarge` on all three clouds (`r6i.4xlarge` / `n2-highmem-16` / `Standard_E16s_v5`).

A 16-vCPU node has ~15.8 allocatable after kube-reserved and ~14.9 after the standard DaemonSets (newrelic/datadog/wiz/kube-system ≈ 1 CPU). So a single 15-CPU parquet pod **cannot schedule on a brand-new empty node of that type**. Cluster-autoscaler correctly declines (`NotTriggerScaleUp: Insufficient cpu`), and because the deployment is `maxSurge=1, maxUnavailable=0`, the surge pod goes `Pending` while the old pods keep serving — the rollout wedges indefinitely and the install looks healthy while silently stuck. (Raising `max_nodes` — core#47476 — can't help; it's a per-node-fit problem, not a ceiling.)

The chart comment "reserve an entire node for parquet" only ever "reserved" a node by being coincidentally 1 CPU short of fitting; the moment per-node overhead crosses ~0.8 CPU (e.g. a newrelic DaemonSet rollout) it flips to permanently unschedulable.

## What

Lower the `parquet` and `parquet-metadata-cache` **request** to **13 CPU** at `medium`+ (limit stays **15** for burst headroom). 13 = ~14.9 schedulable − ~1.8 headroom, so a replica fits the default 16-vCPU node on all three clouds with margin for DaemonSet growth, and the wedged rollout unblocks (the surge pod now fits, so autoscaler will add a node when needed).

Dropping `request` below `limit` moves parquet to Burstable QoS — intentional; it can still burst to 15. No anti-affinity added: the chart already applies default **soft** `topologySpreadConstraints` on `kubernetes.io/hostname`, which spreads replicas one-per-node without hard-blocking scheduling. On 16-vCPU nodes two 13-CPU replicas can't co-locate anyway (26 > 15.8), so one-per-node still holds.

## Validation

- `helm lint` passes; rendering at `global.size=medium` yields parquet `requests.cpu: "13"` / `limits.cpu: "15"` with the existing soft hostname/zone `topologySpreadConstraints` intact.
- Grounded in a live scan of 102 AWS single-tenant clusters: **30** run the 15-CPU profile on 16-vCPU nodes (each on ≤0.8 CPU of headroom), and **3 are already wedged** with a `Pending` parquet pod (`shadybrook`, `pinterest`, `us1-staging-dog`). `small` is unaffected (parquet requests 1 CPU there) and untouched here.

## Rollout / follow-ups

- This chart edit reaches installs only via a chart release + the deployer advancing `operator_chart_version` (or the release channel). The immediate lever for the 3 currently-wedged installs is a per-install CR `parquet.resources.requests.cpu: "13"` override (operator precedence CR > defaults).
- Follow-up 1: a deploy-time schedulability preflight (fail fast when a single-pod request exceeds node allocatable) so the two ladders can't silently drift again — spans this repo + the deployer.
- Follow-up 2: verify api/frontend inherit the default PDB as intended.

Ticket: https://coreweave.atlassian.net/browse/WB-36925

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated the Helm chart version to `0.44.0`.
* **Changes**
  * Reduced CPU request sizing for medium, large, xlarge, and xxlarge tiers for parquet-related components, while keeping CPU limits unchanged.
  * Adjusted sizing notes to match the updated resource requests.

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.43.2
+version: 0.44.0
 appVersion: 1.0.0
 icon: https://wandb.ai/logo.svg
 
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -2711,7 +2711,7 @@ parquet:
           cpu: "15"
           memory: 64Gi
         requests:
-          cpu: "15"
+          cpu: "13"
           memory: 64Gi
       autoscaling:
         horizontal:
@@ -2724,21 +2724,22 @@ parquet:
           cpu: "15"
           memory: 64Gi
         requests:
-          cpu: "15"
+          cpu: "13"
           memory: 64Gi
       autoscaling:
         horizontal:
           enabled: true
           maxReplicas: 2
           minReplicas: 2
-    # At larger t-shirt sizes, reserve an entire node for parquet.
+    # medium+ request (13 CPU) is sized to fit the default 16-vCPU node pool
+    # after DaemonSet/kube-reserved overhead; limit stays 15 CPU for burst.
     xlarge:
       resources:
         limits:
           cpu: "15"
           memory: 64Gi
         requests:
-          cpu: "15"
+          cpu: "13"
           memory: 64Gi
       autoscaling:
         horizontal:
@@ -2751,7 +2752,7 @@ parquet:
           cpu: "15"
           memory: 64Gi
         requests:
-          cpu: "15"
+          cpu: "13"
           memory: 64Gi
       autoscaling:
         horizontal:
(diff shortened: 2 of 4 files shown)


j7m4 (approved):

zacharyblasczyk:
Going to pause this.

PR #641, opened 2026-07-08
chore: add identity db migration prehook

adityachoudhari26:
## Summary by CodeRabbit

* **Chores**
  * Updated the orchestrator Helm chart version to **1.3.1**.
  * Improved database migration job handling during installs and upgrades, including updated hook behavior and cleanup settings.
  * Added a new identity migration job with resource settings and safer startup behavior.
  * Adjusted the workspace migration job configuration to use the updated hook setup.

### DIFF
--- a/charts/orchestrator/values.yaml
+++ b/charts/orchestrator/values.yaml
@@ -191,21 +191,12 @@ identity:
     repository: orchestrator/identity
     tag: latest
 
-  # Run the auth DB (Drizzle better-auth) migrations before the identity
-  # Deployment rolls. helmHook only annotates jobs, so the Deployment/Service
-  # render normally.
-  helmHook:
-    enabled: true
-    hook: "pre-install,pre-upgrade"
-    hookWeight: "-5"
-    hookDeletePolicy: "before-hook-creation,hook-succeeded"
-
-  # The job container inherits the identity image (same tag as the deployment) and
-  # AUTH_DATABASE_URL from envTpls below via wandb-base. It runs the same compiled
-  # migrate the old dedicated migrations image ran. SKIP_AUTH_DB_CREATE assumes the
-  # auth DB is pre-provisioned (like the engine assumes its DB exists).
   jobs:
     auth-migrate:
+      annotations:
+        "helm.sh/hook": "pre-install,pre-upgrade"
+        "helm.sh/hook-weight": "-5"
+        "helm.sh/hook-delete-policy": "before-hook-creation,hook-succeeded"
       backoffLimit: 3
       ttlSecondsAfterFinished: 300
       containers:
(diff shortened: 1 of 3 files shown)


jsingleton-dev (approved):

charts/orchestrator/values.yaml
```diff
@@ -191,6 +191,36 @@ identity:
...
+  # AUTH_DATABASE_URL from envTpls below via wandb-base. It runs the same compiled
+  # migrate the old dedicated migrations image ran. SKIP_AUTH_DB_CREATE assumes the
+  # auth DB is pre-provisioned (like the engine assumes its DB exists).
+  jobs:
+    auth-migrate:
+      backoffLimit: 3
+      ttlSecondsAfterFinished: 300
+      containers:
+        auth-migrate:
+          command: ["node", "/app/packages/db/dist/migrate.js"]
```
zacharyblasczyk, line 204: nit: this only gates the rollout if `migrate.js` exits non-zero on failure. As written in ctrlplane it does `.catch(console.error)` with no `process.exit`, so a failed auth migration still exits 0 → the Job is marked Succeeded and this pre-upgrade hook passes, and the release proceeds on an un-migrated DB. (Fixing in ctrlplane `packages/db/migrate.ts`.)

charts/orchestrator/values.yaml
```diff
@@ -191,6 +191,36 @@ identity:
...
     tag: latest
 
+  # Run the auth DB (Drizzle better-auth) migrations before the identity
+  # Deployment rolls. helmHook only annotates jobs, so the Deployment/Service
+  # render normally.
+  helmHook:
+    enabled: true
+    hook: "pre-install,pre-upgrade"
+    hookWeight: "-5"
+    hookDeletePolicy: "before-hook-creation,hook-succeeded"
```
zacharyblasczyk: nit: enabling `helmHook` also turns the identity **ServiceAccount** into a pre-install/pre-upgrade hook with `hook-succeeded` (see the `…-identity` SA in the snapshot diff). The identity Deployment references that SA as a normal resource, so Helm can delete it after the hook phase. Mirrors the existing workspace-engine pattern, but worth confirming the SA survives for the Deployment.

charts/orchestrator/values.yaml
```diff
@@ -191,6 +191,36 @@ identity:
...
+  # auth DB is pre-provisioned (like the engine assumes its DB exists).
+  jobs:
+    auth-migrate:
+      backoffLimit: 3
+      ttlSecondsAfterFinished: 300
+      containers:
+        auth-migrate:
+          command: ["node", "/app/packages/db/dist/migrate.js"]
+          env:
+            SKIP_AUTH_DB_CREATE: "true"
```
zacharyblasczyk, line 206: nit: with `SKIP_AUTH_DB_CREATE=true` + a fail-loud migrate, a pre-upgrade will now **block** if the auth DB isn't reachable. `global.authDatabase.host`/`password` default to empty (AUTH_DATABASE_URL renders with literal `$(AUTH_DB_HOST)`/`$(AUTH_DB_PASSWORD)` in the default snapshot) — real deployments must set them or upgrades fail here.

zacharyblasczyk (commented):
A few nits on the auth-migrate hook (all reliability, none blocking the plumbing — the image does ship /app/packages/db/dist/migrate.js correctly).

PR #650, opened 2026-07-15
feat(glue): Add glue leaderElection settings

wnevis-cmyk:
- Bumps chart version
- Adds glue leaderElection settings and tests

## Summary by CodeRabbit

* **New Features**
  * Added optional leader election for Glue replicas, with configurable lease name and conditional environment wiring.
  * Added a Helm test configuration for Glue leader election in simple HA mode.
  * Added ClickHouse Keeper readiness gating before running database migrations.

* **Bug Fixes**
  * Reduced ClickHouse OLAP migration races when the global API is enabled.
  * Improved chart-testing coverage and snapshot-build reliability.

* **Chores**
  * Bumped the Helm chart version to `0.43.6`.

* **Tests**
  * Updated Helm test configurations for more predictable single-node OLAP setup and disabled app component in that scenario.


test-configs/operator-wandb/__snapshots__/snap-ch-migration-job.snap
```diff
@@ -5077,269 +5096,6 @@ spec:
         - mountPath: /tmp/
           name: temp-dir
       initContainers:
-      - name: migrate-db
```
j7m4, line 5080: since leader election allows all initContainers to run, I assume there is something on the glue-side to check if the running process is the "leader"

j7m4 (approved):

test-configs/operator-wandb/__snapshots__/snap-ch-migration-job.snap
wnevis-cmyk: There is, yeah, that's kind of the core piece of it

PR #672, opened 2026-08-03
fix(weave): share release-derived trace service account

nikumar1206:
## Summary

- Derive the shared Weave Trace ServiceAccount from the Helm release name.
- Have the trace workers reuse that account without creating component accounts.
- Keep Azure workload-identity account selection unchanged.
- Leave the separate base `weave` service unchanged.

## Summary by CodeRabbit

- **New Features**
  - Added configurable shared Weave Trace identity support for the main trace service and worker services.
  - Automatically derives a release-scoped service account name and supports custom annotations.
  - Prevents duplicate service account creation when shared identity is enabled.
  - Shared identity can be enabled or disabled through chart values.

- **Improvements**
  - Updated Helm chart versions and bundled components to compatible releases.

- **Tests**
  - Added coverage for service account naming, annotations, enablement, and worker deployment behavior.

## Validation (live cluster, 2026-08-19)

Tested end-to-end on a fresh EKS cluster, simulating the managed-install condition: a full `operator-wandb` install (release name `wandb`) with Gorilla's `app.internalJWTMap` overridden to trust **only** `system:serviceaccount:<ns>:wandb-weave-trace` (i.e. the Terraform-supplied list that wholesale-replaces the chart default).

**Bug reproduced on released 0.44.5:**
- Chart created per-workload SAs (`wandb-weave-trace`, `-trace-worker`, `-evaluate-model-worker`).
- Trace worker's dangerzone call (`GET /service-dangerzone/secrets/...` with its projected SA JWT in `Gorilla-Internal-JWT`) → **`401 {"error":"unknown subject"}`**.
- Trace server (trusted subject) → authorized. Exact production failure.

**Upgraded in place to this PR's build (0.44.6):**
- `helm upgrade` completed in one revision — no immutable-field errors, no hooks failed, no manual steps, **no Gorilla config change or restart**.
- Per-worker SA objects removed; trace worker + evaluate-model worker now run as `wandb-weave-trace`; `wandb-weave` untouched.
- Same dangerzone call from the worker → **authorized** (404 for a nonexistent entity — identical response to the trusted trace server). 401 gone against the unchanged JWT map.

**Functional check post-upgrade:** admin login + GraphQL `viewer` OK; weave SDK session logged 10 nested `@weave.op` calls and read them back correctly (`get_calls`) — write and read paths healthy under the consolidated identity.

### Notes for reviewers
1. `useWeaveTraceIdentity` takes precedence over `serviceAccount.create`/`name` (only the Azure path wins over it). Since it defaults to `true` for the workers, installs that set a custom SA name/annotations (e.g. IRSA) on a worker will silently lose that SA on upgrade unless they set `useWeaveTraceIdentity: false`. Worth a values comment / changelog line.
2. Pre-existing (not this PR): `weave-evaluate-model-worker` and `weave-trace-agent-scoring-worker` have no `weave-trace-internal-jwt` projected volume in values — they can't make JWT-authed dangerzone calls until that's added. This consolidation makes adding it trivially safe (same trusted subject).
3. Coverage: trace-worker and evaluate-model-worker validated live; agent-scoring-worker not exercised (`install: false` pending 0.82) but uses the identical values mechanism.

### Update (2026-08-21): umbrella-owned SA + JWT map cleanup

- The shared `<release>-weave-trace` SA is now explicitly owned by the umbrella chart (`templates/weave-trace-serviceaccount.yaml`, same pattern as `azure-storage-serviceaccount.yaml`/`wandb-bucket-access`), with `weave-trace` flipped to `useWeaveTraceIdentity: true` so all four workloads consume the flag uniformly — no implicit owner, no footgun if the flag is toggled on `weave-trace`. `weave-trace.serviceAccount.annotations` pass through to the umbrella SA (needed for IRSA/Workload-Identity users).
- Default `internalJWTMap` drops the dead `-weave-trace-worker` subject and derives the surviving entry from the shared `wandb.weaveTraceServiceAccountName` helper (evaluated via the existing `tpl` wrap in `api.yaml`), so the trust map and the SA object share one source of truth.
- Tests: `weave_trace_service_account_test.yaml` extended (umbrella render, annotations passthrough, disabled-gate, subchart no longer renders the SA) — 11/11 pass, full operator-wandb suite 60/60. Snapshots regenerated; the `weave-trace` SA's owning template change is visible in the snap diffs, rendered names and JWT subject strings unchanged, so the live validation above still holds.

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -853,9 +853,7 @@ app:
     tag: latest
   # TODO(dpanzella) Read from here for backwards compatability, needs to be moved so API can access as well
   internalJWTMap:
-    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace"
-      issuer: "https://kubernetes.default.svc.cluster.local"
-    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace-worker"
+    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ include `wandb.weaveTraceServiceAccountName` . }}"
       issuer: "https://kubernetes.default.svc.cluster.local"
   env:
     GORILLA_LICENSE_CERT_PATH: "/var/app/jwks.json"
@@ -3856,6 +3854,8 @@ mcp-server:
 
 weave-trace:
   install: false
+  serviceAccount:
+    useWeaveTraceIdentity: true
   deploymentPostfix: "bc"
   podLabels:
     azure.workload.identity/use: '{{ include `wandb.weaveTraceUsesAzureWorkloadIdentity` . }}'
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -853,6 +853,8 @@ app:
     tag: latest
   # TODO(dpanzella) Read from here for backwards compatability, needs to be moved so API can access as well
   internalJWTMap:
+    - subject: "system:serviceaccount:{{ .Release.Namespace }}:wandb-weave-trace"
+      issuer: "https://kubernetes.default.svc.cluster.local"
     - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace"
       issuer: "https://kubernetes.default.svc.cluster.local"
     - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace-worker"
@@ -3856,6 +3858,9 @@ mcp-server:
 
 weave-trace:
   install: false
+  serviceAccount:
+    create: true
+    name: wandb-weave-trace
   deploymentPostfix: "bc"
   podLabels:
     azure.workload.identity/use: '{{ include `wandb.weaveTraceUsesAzureWorkloadIdentity` . }}'
@@ -3997,6 +4002,9 @@ weave-trace:
 
 weave-trace-worker:
   install: false
+  serviceAccount:
+    create: false
+    name: wandb-weave-trace
   deploymentPostfix: "bc"
   podLabels:
     azure.workload.identity/use: '{{ include `wandb.weaveTraceUsesAzureWorkloadIdentity` . }}'
@@ -4109,6 +4117,9 @@ weave-trace-worker:
 
 weave-evaluate-model-worker:
   install: false
+  serviceAccount:
+    create: false
+    name: wandb-weave-trace
   deploymentPostfix: "bc"
   podLabels:
     azure.workload.identity/use: '{{ include `wandb.weaveTraceUsesAzureWorkloadIdentity` . }}'
@@ -4215,6 +4226,9 @@ weave-evaluate-model-worker:
 
 weave-trace-agent-scoring-worker:
   install: false
+  serviceAccount:
+    create: false
+    name: wandb-weave-trace
   deploymentPostfix: "bc"
   podLabels:
     azure.workload.identity/use: '{{ include `wandb.weaveTraceUsesAzureWorkloadIdentity` . }}'
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -778,11 +778,10 @@ app:
   image:
     repository: wandb/local
     tag: latest
-  # TODO(dpanzella) Read from here for backwards compatability, needs to be moved so API can access as well
+  # One shared Weave subject keeps per-install issuer overrides from clobbering
+  # the identity used by additional Weave workloads.
   internalJWTMap:
-    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace"
-      issuer: "https://kubernetes.default.svc.cluster.local"
-    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace-worker"
+    - subject: "system:serviceaccount:{{ .Release.Namespace }}:wandb-weave"
       issuer: "https://kubernetes.default.svc.cluster.local"
   env:
     GORILLA_LICENSE_CERT_PATH: "/var/app/jwks.json"
@@ -3105,6 +3104,11 @@ clickhouseMigrationJob:
 
 weave:
   install: true
+  # All Weave workloads use this ServiceAccount unless their own serviceAccount
+  # values are overridden. It carries the cloud workload identity once.
+  serviceAccount:
+    create: true
+    name: wandb-weave
   deploymentPostfix: "bc"
   cache:
     # Keep WEAVE_CACHE_DURATION_DAYS unset or 0 for size-triggered cleanup.
@@ -3443,6 +3447,9 @@ mcp-server:
 
 weave-trace:
   install: false
+  serviceAccount:
+    create: false
+    name: wandb-weave
   deploymentPostfix: "bc"
   image:
     repository: wandb/weave-trace
@@ -3581,6 +3588,9 @@ weave-trace:
 
 weave-trace-worker:
   install: false
+  serviceAccount:
+    create: false
+    name: wandb-weave
   deploymentPostfix: "bc"
   image:
     repository: wandb/weave-trace
@@ -3690,6 +3700,9 @@ weave-trace-worker:
 
 weave-evaluate-model-worker:
   install: false
+  serviceAccount:
+    create: false
+    name: wandb-weave
   deploymentPostfix: "bc"
   image:
     repository: wandb/weave-trace
@@ -3793,6 +3806,9 @@ weave-evaluate-model-worker:
 
 weave-trace-agent-scoring-worker:
   install: false
+  serviceAccount:
+    create: false
+    name: wandb-weave
   deploymentPostfix: "bc"
   image:
     repository: wandb/weave-trace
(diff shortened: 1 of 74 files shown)


charts/operator-wandb/values.yaml
```diff
@@ -853,6 +853,8 @@ app:
     tag: latest
   # TODO(dpanzella) Read from here for backwards compatability, needs to be moved so API can access as well
   internalJWTMap:
+    - subject: "system:serviceaccount:{{ .Release.Namespace }}:wandb-weave-trace"
+      issuer: "https://kubernetes.default.svc.cluster.local"
     - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace"
```
amwarrier: lets use this SA itself. `.Release.Name` will be wandb.

amwarrier:
Confirmed how the four Weave-trace subcharts currently get their ServiceAccount: they're all aliases of `wandb-base` (see `charts/operator-wandb/Chart.yaml:40-64`) and resolve their SA through `wandb-base.serviceAccountName` in `charts/wandb-base/templates/_helpers.tpl:87-95`, which defaults to `{{ .Release.Name }}-<alias>` via `wandb-base.fullname`. With `.Release.Name = wandb`, `weave-trace` already produces `wandb-weave-trace` today — so the new `weave-trace.serviceAccount` block in this PR just restates the default and can be dropped.

The bigger issue: the three worker aliases hardcode `name: wandb-weave-trace` as a literal string. `.Values.serviceAccount.name` is not `tpl`'d by the helper, so `'{{ .Release.Name }}-weave-trace'` won't work as a values-side fix, and the current literal breaks the moment anyone deploys with a release name other than `wandb`.

### Suggested design

Mirror the existing `azureStorageServiceAccountEnabled` / `azureStorageServiceAccountName` pattern already in `charts/wandb-base/templates/_helpers.tpl` — that helper is generic (backs every `wandb-base`-aliased subchart, not Weave-specific), and it already has one shared-SA short-circuit branch. Add a second one for Weave-trace:

**1. In `charts/wandb-base/templates/_helpers.tpl`**, add a new short-circuit branch and companion helper:

```gotemplate
{{- define "wandb-base.serviceAccountName" -}}
  {{- if include "wandb-base.azureStorageServiceAccountEnabled" . | trim | eq "true" }}
{{- include "wandb-base.azureStorageServiceAccountName" . }}
  {{- else if .Values.serviceAccount.useWeaveTraceIdentity }}
{{- include "wandb-base.weaveTraceServiceAccountName" . }}
  {{- else if .Values.serviceAccount.create }}
{{- default (include "wandb-base.fullname" .) .Values.serviceAccount.name }}
  {{- else }}
{{- default "default" .Values.serviceAccount.name }}
  {{- end }}
{{- end }}

{{- define "wandb-base.weaveTraceServiceAccountName" -}}
{{- printf "%s-weave-trace" .Release.Name -}}
{{- end }}
```

**2. In `charts/wandb-base/templates/serviceaccount.yaml`**, extend the guard so opt-in workloads don't create their own SA:

```gotemplate
{{- if and .Values.serviceAccount.create
      (not (include "wandb-base.azureStorageServiceAccountEnabled" . | trim | eq "true"))
      (not .Values.serviceAccount.useWeaveTraceIdentity) -}}
```

**3. In `charts/wandb-base/values.yaml`**, add the flag default:

```yaml
serviceAccount:
  useWeaveTraceIdentity: false
```

**4. In `charts/operator-wandb/values.yaml`**, drop the `weave-trace.serviceAccount` block entirely, and on each of `weave-trace-worker`, `weave-evaluate-model-worker`, `weave-trace-agent-scoring-worker` set only:

```yaml
serviceAccount:
  useWeaveTraceIdentity: true
```

No hardcoded `name`, no `create: false` — the helper handles both, and it stays correct regardless of `.Release.Name`.

### Follow-up (not required for this PR)

Once all workers are on the shared SA, the three per-worker entries in `internalJWTMap` (`-weave-trace-worker`, `-weave-evaluate-model-worker`, `-weave-trace-agent-scoring-worker`) will be unused and can be removed. Keeping them during rollout is fine.

charts/operator-wandb/values.yaml
nikumar1206: Implemented the release-derived helper approach: the trace workers opt into `useWeaveTraceIdentity`, which selects `<release>-weave-trace` and suppresses their component accounts. Added non-`wandb` release coverage.

amwarrier:
Follow-up on my previous comment — design landed cleanly, but seeing it merged I want to revisit **who owns the SA object**. Today `weave-trace` is the implicit owner via `wandb-base.fullname` happening to equal `<release>-weave-trace`; a reader who "helpfully" adds `useWeaveTraceIdentity: true` to `weave-trace` for symmetry would silently kill the SA and leave four pods `ServiceAccountNotFound`. Worth closing that footgun.

Precedent for the fix is exact: [`charts/operator-wandb/templates/azure-storage-serviceaccount.yaml`](https://github.com/wandb/helm-charts/blob/main/charts/operator-wandb/templates/azure-storage-serviceaccount.yaml) already owns `wandb-bucket-access` at the umbrella level for the same reason — a shared identity that many W&B workloads converge on. The umbrella is the only file location that renders once per release; the `weave-*` workloads are all `wandb-base` aliases with no dedicated subchart to hold this. Filename carries the scope.

### Concretely

**1. Helper in `charts/operator-wandb/templates/_bucket.tpl`** (next to `weaveTraceUsesAzureWorkloadIdentity`):
```gotemplate
{{- define "wandb.weaveTraceServiceAccountName" -}}
  {{- printf "%s-weave-trace" .Release.Name -}}
{{- end }}
```

**2. New umbrella template `charts/operator-wandb/templates/weave-trace-serviceaccount.yaml`** — gate only on `weave-trace.install`, otherwise render:
```gotemplate
{{- if (index .Values "weave-trace" "install") }}
apiVersion: v1
kind: ServiceAccount
metadata:
  name: {{ include "wandb.weaveTraceServiceAccountName" . }}
  labels:
    {{- include "wandb.commonLabels" . | nindent 4 }}
{{- end }}
```

**3. Flip `weave-trace` to `useWeaveTraceIdentity: true` in `charts/operator-wandb/values.yaml`** so all four subcharts consume the flag uniformly. No implicit owner.

**4. `wandb-base.weaveTraceServiceAccountName` stays as-is** — the two `printf`s are load-bearingly identical strings; a one-line comment on each pointing at the other is enough.

### Testing

- Extend `weave_trace_service_account_test.yaml` with a case rendering the new umbrella template and asserting `metadata.name: custom-release-weave-trace`.
- Regenerate `weave-trace-with-worker.snap`: same `chartsnap-weave-trace` SA, new owner (`operator-wandb/templates/weave-trace-serviceaccount.yaml`). That diff is the receipt.

Live re-test isn't needed — the JWT map trusts `<release>-weave-trace` and that string doesn't change.

### JWT map cleanup (do it in this PR)

`charts/operator-wandb/values.yaml:855-859` still ships two default trusted subjects:
```yaml
app:
  internalJWTMap:
    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace"
      issuer: "https://kubernetes.default.svc.cluster.local"
    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ .Release.Name }}-weave-trace-worker"
      issuer: "https://kubernetes.default.svc.cluster.local"
```

Two things to fix:

**1. Drop the `-weave-trace-worker` entry.** Post-consolidation no pod runs as that subject; the entry is dead. (For managed installs whose Terraform wholesale-replaces `app.internalJWTMap` — including the one you tested against — this is cosmetic; but the chart default is what OSS / self-hosted users get, and they'd wonder why a `-weave-trace-worker` subject is trusted for a pod that no longer exists.)

**2. Make the remaining entry reference the umbrella SA symbolically** so the JWT map and the SA object share one source of truth:
```yaml
app:
  internalJWTMap:
    - subject: "system:serviceaccount:{{ .Release.Namespace }}:{{ include `wandb.weaveTraceServiceAccountName` . }}"
      issuer: "https://kubernetes.default.svc.cluster.local"
```
This works because `api.yaml:48` already wraps the rendered map in `tpl (...)`, so the `include` is evaluated on the second pass just like the existing `{{ .Release.Namespace }}` interpolation. If we ever rename the helper's output, the JWT map follows automatically instead of going stale.

Worth noting for the incident record: the chart default *never* trusted `-weave-evaluate-model-worker` or `-weave-trace-agent-scoring-worker` — those two workers would have hit 401 against a stock install before this PR, since they had their own component SAs that were absent from the map. This PR fixes that as a side effect by routing all four onto `-weave-trace`.

### Follow-up in wandb/core (do it together, or immediately after)

All three cloud single-tenant TF modules wholesale-replace `app.internalJWTMap` — they don't rely on the chart default, so **managed installs continue working with this PR alone**. But every module still ships the now-dead `-weave-trace-worker` subject and, on AWS/Azure, an unused SA-name local. Clean those up in a `wandb/core` PR:

- **AWS** — [`services/deployer/terraform/deploy/aws/single-tenant/module/main.tf:284-286, 393-404`](https://github.com/wandb/core/blob/master/services/deployer/terraform/deploy/aws/single-tenant/module/main.tf#L284): drop `local.weave_trace_worker_sa_name` and the second `internalJWTMap` entry that references it.
- **Azure** — [`services/deployer/terraform/deploy/azure/single-tenant/module/main.tf:428-429, 531-540`](https://github.com/wandb/core/blob/master/services/deployer/terraform/deploy/azure/single-tenant/module/main.tf#L428): drop `local.weave_trace_worker_service_account_name` and the second `internalJWTMap` entry.
- **GCP** — [`services/deployer/terraform/deploy/google/single-tenant/module/main.tf:39, 504-513`](https://github.com/wandb/core/blob/master/services/deployer/terraform/deploy/google/single-tenant/module/main.tf#L39): drop `local.k8s_sa_map.weave_trace_worker` and the second `internalJWTMap` entry.

amwarrier (approved):

jsingleton-dev (approved):

PR #673, opened 2026-08-06
feat(backend): enable bufstream in MI filemeta

elsteveogrande:
https://coreweave.atlassian.net/browse/WB-38269

Allows `filemeta` to read file metadata straight from the run-updates kafka topic.  Doesn't change behavior immediately, but allows this to be eventually turned on with statsig (`filemeta_run_updates_source_deployment`, `filemeta_disable_spout_deployment`).

The run-updates source for file metadata is consumed only when the
`filemeta_run_updates_source_deployment` Statsig gate is enabled;
otherwise filemeta falls back to `GORILLA_FILE_METADATA_SOURCE` (redis).

Tested on `runsv2-test` with latest gorilla changes + these user spec overrides:

```
  filemeta:
    image:
      tag: 0.84.0-daily.14
    pubSub:
      subscription: ''
    kafka:
      consumerGroupId: filemeta
    env:
      GORILLA_RUN_UPDATE_SHADOW_QUEUE_SUBSCRIPTIONS_FILEMETA: >-
        kafka://bufstream.bufstream.svc.cluster.local:9092/wandb-run-updates-shadow?consumer_group_id=filemeta&first_boot_lookback=1h
      GORILLA_FILE_METADATA_SOURCE: noop://
```

(which _approximately_ equals the change in this PR's first commit)

## Summary by CodeRabbit

* **New Features**
  * Added configuration for processing run updates through a Pub/Sub subscription and dedicated Kafka consumer group.
  * Enabled a shared queue environment configuration and imported Kafka connection settings.
  * Added support for a run-update shadow queue.

### DIFF
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -1307,8 +1307,18 @@ filemeta:
   image:
     repository: wandb/megabinary
     tag: latest
+  pubSub:
+    # pubSub subscription comes from terraform
+    subscription: ""
+  kafka:
+    # unique consumer group so it reads the run-update shadow topic independently
+    consumerGroupId: "filemeta"
+  env:
+    GORILLA_RUN_UPDATE_SHADOW_QUEUE_SUBSCRIPTIONS_FILEMETA:
+      value: '{{ include "wandb.runUpdateShadowQueue" . }}'
   envFrom:
     "{{ .Release.Name }}-bucket-configmap": "configMapRef"
+    "{{ .Release.Name }}-kafka-configmap": "configMapRef"
     "{{ .Release.Name }}-redis-configmap": "configMapRef"
     "{{ .Release.Name }}-global-configmap": "configMapRef"
     "{{ .Release.Name }}-global-secret": "secretRef"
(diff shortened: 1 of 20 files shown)


charts/operator-wandb/values.yaml
```diff
@@ -1307,8 +1307,21 @@ filemeta:
...
     repository: wandb/megabinary
     tag: latest
+  pubSub:
+    # pubSub subscription comes from terraform
+    subscription: ""
+  kafka:
+    # unique consumer group so it reads the run-update shadow topic independently
+    consumerGroupId: "filemeta"
+  env:
+    # The run-updates source for file metadata: consumed only when the
```
jessicaxiang: this comment may not be needed since we're going to have to remember to remove this once the statsig flag no longer applies. I think this comment makes more sense in the PR description than in the code.

elsteveogrande: :+1:

charts/operator-wandb/values.yaml
```diff
@@ -1307,8 +1307,21 @@ filemeta:
...
+    # filemeta_run_updates_source_deployment Statsig gate is enabled;
+    # otherwise filemeta falls back to GORILLA_FILE_METADATA_SOURCE (redis).
```
elsteveogrande:
```suggestion
```

zacharyblasczyk (commented):
This looks correct to me. The env only makes the run-updates subscription available; it does not enable the new source. Activation still requires a megabinary version containing the filemeta run-updates implementation, plus `filemeta_run_updates_source_deployment` enabled and a filemeta restart. With the gate off, it continues using `GORILLA_FILE_METADATA_SOURCE`.

Minor pre-existing concern: `wandb.runUpdateShadowQueue` hardcodes port 9092 instead of using `$(KAFKA_BROKER_PORT)`. That is shared existing behavior, not introduced by this PR, and is fine if MI/Bufstream always uses 9092.

zacharyblasczyk (approved):

PR #681, opened 2026-08-17
fix: drop Linux capabilities by default

jsingleton-dev:
## What

- Drop all Linux capabilities by default in `wandb-base` container security contexts.
- Bump `wandb-base` to 0.12.7 and update the exact dependency pins for `operator-wandb`, Lumen, and orchestrator.
- Regenerate dependency locks and rendered snapshots.

## Why

The Kubernetes Restricted Pod Security Standard requires containers to drop `ALL` capabilities. `allowPrivilegeEscalation: false` prevents gaining new privileges but does not explicitly remove capabilities granted at container startup.

This addresses the Kyverno restricted-policy capability findings for workloads that inherit the `wandb-base` security context.

## CIS coverage

This directly addresses these capability-specific controls reported through Kyverno policy `westest-cis-pod-security-restricted/restricted`:

- CIS 5.2.8 — Minimize the admission of containers with the `NET_RAW` capability (`capabilities_baseline`).
- CIS 5.2.9 — Minimize the admission of containers with capabilities assigned (`capabilities_restricted`).

Kyverno reports the Restricted Pod Security Standard as one aggregate rule. A workload will continue to fail that rule if it has another Restricted PSS violation, such as privileged mode, host namespaces, `allowPrivilegeEscalation`, root execution, or seccomp configuration.

This does not directly clear a capability-specific kube-bench finding in the supplied results. It contributes to GKE's aggregate PSS Baseline-or-stricter control, but does not satisfy that control by itself.

## Impact

Containers and init containers using the default `wandb-base` security context will render:

```yaml
capabilities:
  add: []
  drop:
    - ALL
```

Stock W&B services already run non-root and use unprivileged ports, so no functional change is expected. The chart should still be rolled through QA before wider Dedicated promotion, with particular attention to migration jobs and custom-CA startup.

## Testing

- `python3 scripts/format_templates.py`
- `python3 -m unittest discover -s scripts/tests`
- `python3 scripts/check_template_maintainability.py`
- `ct lint` for `wandb-base`, `operator-wandb`, Lumen, and orchestrator
- Snapshot suites for all four affected charts
- `helm unittest ./charts/operator-wandb/` — 49 tests passed
- Focused render assertion confirming main and init containers inherit `drop: [ALL]`
- `git diff --check`

### DIFF
--- a/charts/lumen/Chart.lock
+++ b/charts/lumen/Chart.lock
@@ -1,9 +1,9 @@
 dependencies:
 - name: wandb-base
   repository: file://../wandb-base
-  version: 0.12.6
+  version: 0.12.7
 - name: wandb-base
   repository: file://../wandb-base
-  version: 0.12.6
-digest: sha256:60d1023fe307cba27e2dfe573be514601dc0dcacb6d72f91323615fedaa42772
-generated: "2026-07-29T18:00:40.632395-05:00"
+  version: 0.12.7
+digest: sha256:6a4d35030d930a8ad4f6a1ada246116749584e084a7a6f2b2f2f48de13e0c990
+generated: "2026-08-17T13:29:39.702668-05:00"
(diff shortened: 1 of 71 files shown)


jsingleton-dev:
Will reopen this later

PR #690, opened 2026-09-01
feat(operator-wandb): split Weave migrator credentials [WB-36557]

zacharyblasczyk:
## Summary

- add an opt-in `global.olap.weaveTrace.migrator` credential path
- override only `WF_CLICKHOUSE_USER` and `WF_CLICKHOUSE_PASS` in the
  `weave-trace-migrate` init container
- require both migrator fields to use Kubernetes `valueFrom` references
- keep runtime Weave containers on the standard Weave credential
- test the stable ESO Secret names produced by `wandb/deployments#858`

The matching `weave` and `weave_migrator` ClickHouse identities are provisioned
by [wandb/deployments#858](https://github.com/wandb/deployments/pull/858). ESO
materializes their GSM credentials as the stable `clickhouse-weave` and
`clickhouse-weave-migrator` Kubernetes Secrets.

During A/B rotation, Delivery Tooling publishes both new GSM versions, confirms
ESO synchronized both Kubernetes Secrets, and rolls the Weave workloads. The
runtime and migrator credentials move together even though their stable Secret
names do not change.

## Safety

- disabled by default
- active only when the explicit Weave OLAP source is selected
- legacy and bundled ClickHouse behavior is unchanged
- inline migrator usernames and passwords are rejected
- runtime and migration credential separation is covered by render tests

## Validation

- `python3 scripts/format_templates.py`
- `python3 -m unittest discover -s scripts/tests` — 16 passing
- `python3 scripts/check_template_maintainability.py`
- `helm lint charts/operator-wandb`
- `helm lint charts/wandb-base`
- `helm unittest charts/operator-wandb` — 75 passing
- operator-wandb snapshots with CI-pinned Helm 3.20.1 — passing
- wandb-base snapshots with CI-pinned Helm 3.20.1 — passing
- full render confirms runtime uses `clickhouse-weave` and only the migration
  init container uses `clickhouse-weave-migrator`

## Summary by CodeRabbit

* **New Features**
  * Added optional, separate Kubernetes secret references for Weave Trace migration credentials.
  * Migration credentials are used only by the migration container; runtime containers continue using standard credentials.
  * Added support for container-specific environment variable configuration.

* **Bug Fixes**
  * Helm rendering now reports clear errors when migration credentials are not provided through supported secret references.

* **Documentation**
  * Documented credential configuration and guidance for coordinating credentials during A/B password rotation.

### DIFF
--- a/charts/lumen/Chart.lock
+++ b/charts/lumen/Chart.lock
@@ -1,9 +1,9 @@
 dependencies:
 - name: wandb-base
   repository: file://../wandb-base
-  version: 0.12.7
+  version: 0.12.8
 - name: wandb-base
   repository: file://../wandb-base
-  version: 0.12.7
-digest: sha256:6a4d35030d930a8ad4f6a1ada246116749584e084a7a6f2b2f2f48de13e0c990
-generated: "2026-08-11T14:59:58.54023-07:00"
+  version: 0.12.8
+digest: sha256:541b57c62629d87b2005d19134e0fba306517057cc9bca65035ecb63b5f2d81d
+generated: "2026-09-02T16:15:46.187432-05:00"
--- a/charts/lumen/Chart.yaml
+++ b/charts/lumen/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: lumen
 description: A Helm chart for deploying the W&B Lumen processor and notebook
 type: application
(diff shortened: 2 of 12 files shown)


dacbd (approved):
might add a snapshot that uses the new field.

PR #691, opened 2026-09-01
fix(operator-wandb): declare nginx log source on the frontend pods

mwhooker:
## What

Adds one autodiscovery annotation to the frontend pods so their nginx logs arrive in Datadog as
`source:nginx` / `service:frontend`:

```yaml
frontend:
  podAnnotations:
    ad.datadoghq.com/frontend.logs: '[{"source": "nginx", "service": "frontend"}]'
```

With `source:nginx`, Datadog's built-in Nginx integration pipeline parses these logs — grok, `level`
extraction, status remapping — so nginx `[warn]` lines stop being classified as errors. Chart
version bumped 0.44.7 → 0.44.8; the snapshot files that render the frontend pod gain exactly these
two lines and nothing else.

## Why

Without an `ad.datadoghq.com` logs annotation, the Datadog Agent infers a container's log source
from its short image name, so these logs arrive as `source:frontend-nginx` — a name that matches no
integration pipeline, leaving the nginx error stream unparsed. The same inference sets the
log-level `service`, overriding the intended `frontend`. Datadog's convention is that `source`
names the technology (which activates integration pipelines and out-of-the-box assets) and
`service` names the app.

Fixing this in pipeline configuration instead does not work: Datadog's Nginx integration pipeline
is read-only (updates are accepted and silently discarded), and cloning its processors into a
custom pipeline would drift whenever Datadog updates the original. Declaring the source at the pod
is the supported mechanism.

## Rollout

Each install picks the annotation up when its operator applies chart ≥ 0.44.8 and the frontend pods
restart. Pods not yet restarted keep emitting the old source and stay unparsed — exactly today's
behavior, so there is no regression window.

Note the effect is deliberately narrow: the frontend's access logs are already emitted as
structured JSON, which the Agent parses before pipelines run, so they are unaffected. Only the
raw-text nginx error-stream lines gain parsing (and correct `warn` severity).

## Testing

`./snapshots.sh run` and `helm unittest` pass locally on CI's toolchain (helm 3.20.1,
cascade-built dependencies with the stakater/prometheus repos). The regenerated snapshots' entire
diff is the two annotation lines across the test configs that install the frontend.

### DIFF
--- a/charts/operator-wandb/Chart.yaml
+++ b/charts/operator-wandb/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: operator-wandb
 description: A Helm chart for deploying W&B to Kubernetes
 type: application
-version: 0.44.7
+version: 0.44.8
 appVersion: 1.0.0
 icon: https://wandb.ai/logo.svg
 
--- a/charts/operator-wandb/values.yaml
+++ b/charts/operator-wandb/values.yaml
@@ -2610,6 +2610,8 @@ history-reader:
 
 frontend:
   install: false
+  podAnnotations:
+    ad.datadoghq.com/frontend.logs: '[{"source": "nginx", "service": "frontend"}]'
   podSecurityContext:
     runAsUser: 101
   image:
--- a/test-configs/operator-wandb/__snapshots__/fmb.snap
+++ b/test-configs/operator-wandb/__snapshots__/fmb.snap
@@ -6090,6 +6090,8 @@ spec:
       app.kubernetes.io/instance: chartsnap
   template:
     metadata:
+      annotations:
+        ad.datadoghq.com/frontend.logs: '[{"source": "nginx", "service": "frontend"}]'
       labels:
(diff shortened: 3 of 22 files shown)


mleonidas (approved):

dacbd (commented):
consider putting this in the managed-spec/release channel since this chart is still publicly used.

jsingleton-dev (approved):

mwhooker:
## Impact review

Before this change we reviewed every consumer of the current `source`/`service` values in the
receiving Datadog org: no monitors, dashboards, log-based metrics, pipelines, archives, or index
routing/exclusion configuration reference them, and `source:nginx` has no existing log population
to collide with, so the rename changes classification only.

The frontend's access logs are structured JSON and are parsed by the Agent before pipelines run, so
they are unaffected; the nginx pipeline's effect lands only on the raw-text error-stream lines,
where `[warn]` records gain correct severity (`[error]` stays error). Rollout is per-pod-restart
with no regression window: pods without the annotation keep today's behavior verbatim.

PR #694, opened 2026-09-01
feat(operator-wandb): gate history updater with OLAP enablement

zacharyblasczyk:
## Summary

- install `history-updater` from the same effective history-store enablement while retaining `history-updater.install` as a compatibility fallback
- make an explicitly set `global.nextGenHistoryStore` authoritative for history OLAP configuration and migrations
- default `history-updater.traceRatio` to `1`
- exercise the enabled updater in the OLAP integration matrix with an in-cluster Bufstream queue
- add precedence and migration coverage, and update snapshots

## Flag precedence

Helm dependency condition lists use the first valid boolean path; they are not an any-true expression. The order is therefore:

1. `global.nextGenHistoryStore` when explicitly set
2. `global.olap.history.enabled` when explicitly set
3. `history-updater.install` as the legacy fallback

The two newer values are documented but omitted from the chart defaults so an unset value can fall through. A YAML anchor is intentionally not used: anchors are resolved while parsing the base values file and do not follow later Helm values overlays, and materializing the anchored boolean would prevent compatibility fallback.

## Validation

- `helm unittest ./charts/operator-wandb/` (77 tests passed)
- `helm lint ./charts/operator-wandb/ --strict`
- `./snapshots.sh run` (all snapshots matched)
- rendered integration includes Bufstream at `bufstream.default.svc.cluster.local` and a dedicated `history-updater` consumer group
- `git diff --check`

### DIFF
--- a/charts/operator-wandb/Chart.lock
+++ b/charts/operator-wandb/Chart.lock
@@ -116,5 +116,5 @@ dependencies:
 - name: clickhouse
   repository: file://charts/clickhouse
   version: 9.1.1
-digest: sha256:1986ad265b0aaca94ee3286ce5b4f08e3841c9813bf22a16cc7ec97925d61a52
-generated: "2026-08-11T14:59:55.991901-07:00"
+digest: sha256:970fbdd2ec25e7a5be166c02778280d56a3b362ea8cfd19d7bafac090b33f9c7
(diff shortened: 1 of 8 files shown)


zacharyblasczyk:
@devin review

adityachoudhari26 (approved):

PR #695, opened 2026-09-02
chore(operator-wandb): increase liveness and readiness probe timeout for anaconda2 in MI

nathancy-wandb:
### What
Increase anaconda2 liveness and readiness probe timeouts from 2s to 120s and 2s to 30s respectively.

### Why
This is to help prevent spurious/premature anaconda2 container restarts when there are many concurrent and high latency calls to anaconda2; this is because the liveness and readiness probes are done via GET requests to /ping, which can take a long time to complete if there are many other requests in the gunicorn queue.

### Testing
Tested anaconda2 before and after the probe change with various artificial delays to anaconda2 endpoints to simulate CPU-bound requests:

[us5.datadoghq.com/notebook/315895/anaconda2-probe-failures-and-restarts?cell-62hptrkm-from_ts=1787678389681&cell-62hptrkm-refresh_mode=paused&cell-62hptrkm-to_ts=1787679689527&cell-9o1qxn3p-from_ts=1786267426744&cell-9o1qxn3p-refresh_mode=paused&cell-9o1qxn3p-to_ts=1786268374399&refresh_mode=paused&from_ts=1786141842983&to_ts=1786145492000](https://us5.datadoghq.com/notebook/315895/anaconda2-probe-failures-and-restarts?cell-62hptrkm-from_ts=1787678389681&cell-62hptrkm-refresh_mode=paused&cell-62hptrkm-to_ts=1787679689527&cell-9o1qxn3p-from_ts=1786267426744&cell-9o1qxn3p-refresh_mode=paused&cell-9o1qxn3p-to_ts=1786268374399&refresh_mode=paused&from_ts=1786141842983&to_ts=1786145492000)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Anaconda2 health-check reliability by extending liveness and readiness probe periods and timeouts.
- **Configuration**
  - Updated Anaconda2 liveness checks to run every 130 seconds with a 120-second timeout.
  - Updated readiness checks to run every 40 seconds with a 30-second timeout.
- **Maintenance**
  - Updated the operator-wandb Helm chart release version.


charts/operator-wandb/values.yaml
```diff
@@ -1484,13 +1484,15 @@ anaconda2:
...
+        periodSeconds: 130
+        timeoutSeconds: 120
         failureThreshold: 5
       readinessProbe:
         httpGet:
           path: /ping
           port: http
-        timeoutSeconds: 2
+        periodSeconds: 40
+        timeoutSeconds: 30
```
jsingleton-dev, line 1495: Keep an eye on your latency and pod saturation after this change goes out. I'm not familiar with anaconda2 but readiness probes often provide a nice back-pressure mechanism on overloaded pods by taking them out of service while they catch up on the the load that caused them to fail the probes. Now that you've increased the time it takes for them to go to unready, it's possible that problem gets worse. Could not be the case for your service just something to keep an eye one.

jsingleton-dev (approved):

charts/operator-wandb/values.yaml
nathancy-wandb: yep great point 

fwiw in the load testing i did, i was still able to observe this backpressure effect. But i'll def take a close watch on the metrics for the next few weeks

PR #707, opened 2026-09-17
fix(operator-wandb): raise ClickHouse history max_open_conns to 500

seongkim0228:
## Summary

The parquet pod's ClickHouse DSN (`GORILLA_HISTORY_STORAGE_ENGINE_ADDRESS`) was being hand-patched with `max_open_conns=500`. Bake it into the chart default:

```yaml
global.olap.history.params:
-  max_open_conns: 300
+  max_open_conns: 500
```

`HISTORY_PARAMS` now renders as `?max_open_conns=500[&...]` for every history consumer (parquet, executor, history-updater, history-reader, CH migration job). `conn_max_lifetime` is intentionally not set: clickhouse-go v2 already defaults it to 1h. No template changes; two snapshots regenerated. Chart bumped to 0.44.20.

## Summary by CodeRabbit

- **Improvements**
  - Increased the default maximum number of open connections for OLAP history data access from 300 to 500.
- **Release**
  - Updated the operator chart version to 0.44.20.


zacharyblasczyk (approved):
Approved, but we should set a good default, and tune elsewhere.

